PR #35, opened 2019-12-10
Don't ignore RefinableDateSingle refiners

cmasonrp:
Check for RefinableDateSingle managed properties in search service


FranckyC:
Thanks @cmasonrp. Good catch.

Issue #16, opened 2019-11-11
Person Field Refiner


gristy58:
Currently person fields in refiners are returned as claims username, is there a way to get it to display the display name, or would this require an enhancement?

FranckyC:
Hi @gristy58 What managed properties are you using for refiners?

gristy58:
I mapped People:Manager to RefinableString

FranckyC:
Can you add this RefinableString to the list of selected properties in the search results and use the debug view? You should see the raw value coming from the search index.

In the refiner WP, value are displayed as is (except for dates). If the raw value doesn't fit your needs, maybe you can try to map an other crawled property.

gristy58:
yeah can you please process claims usernames and display display name? as there is no other way i can figure out to display the name properly.

FranckyC:
I will see what I can do but I'm very busy until the end of the year so this won't be soon :( Maybe @wobba  @YannickRe can help if you have some time?

FranckyC:
Hi @gristy58 . This should be resolved by PR #37. Let me know if it works for you.

FranckyC:
Resolved by #59.

PR #37, opened 2019-12-11
[Refiners] Person Field Refiner #16

PooLP:
A little contribution for issue #16, it needs to be improved later.


FranckyC:
Hi @PooLP, looks great! However, before merging, I would suggest some improvements:
- Instead of making the call directly in the component, I suggest you to create an `UserService` service, following the same pattern as others (with mock class). Services must be instantiated in the main WebPart class and passed through props to child controls. This way, it is easier to mock.
- Don't pass the WebPartContext as prop for the component. What you need is only the `SPHttpClient` instance. For that, you can use SPFx service scopes in your service. Get the service scope form the main WebPart when instantiating your service (in the constructor).

PooLP:
Hi, ok, you're right. I maked this too fast. I would look this weekend.

PooLP:
In refiner panel use, after selected a refiner with specific template (ex : persona) I don't arrive to return my value with the template used to be able to do the formatting, you have any idea to return the template used with selected refiners value ?

For date range you use it :
```ts
// Date refiner value
      if (/range\(.+\)/.test(value.RefinementToken) && value.RefinementName !== value.RefinementValue) {
        filterName = `${filterName} ${value.RefinementValue}`;
      }
```

but for another template, I do not know how to do.

image

FranckyC:
Hi @PooLP. I will look into this ASAP (pretty busy right now). Because it has merge conflicts, it may take some time. Sorry about that but once again, thank you for your help!

PooLP:
It's ok, I can only work on it on small lot of weekends (unfortunately because it's an interesting project).
For conflicts, it's strange, I did not have conflicts this weekend after my last commit, But you may have worked on it too ;).

FranckyC:
Yes I've made some updates regarding the section background support.

PooLP:
It's ok for merge + little modifcation for theme variant support. You can check.

You would have an idea for that ? :

> In refiner panel use, after selected a refiner with specific template (ex : persona) I don't arrive to return my value with the template used to be able to do the formatting.
> Do you have any idea for to return the template used with selected refiners value ?
> 
> For date range you use it :
> 
> ```ts
> // Date refiner value
>       if (/range\(.+\)/.test(value.RefinementToken) && value.RefinementName !== value.RefinementValue) {
>         filterName = `${filterName} ${value.RefinementValue}`;
>       }
> ```
> 
> I have a same like methods for persona, but I would like a few things cleaner.
> 
> image

wobba:
@FranckyC will you handle this one?

FranckyC:
@wobba @PooLP I was on vacation but yes I will take care of this one. This week end or next week.

FranckyC:
@PooLP Can you provide a list of crawled properties (through RefinableStringXX) compatible with your refiner template? I tried with Author but doesn't seem to work.

PooLP:
@FranckyC For authors, you must map ows_q_USER_Author on a refinableString. All ows_q_USER* should work if they are map on a refinableString.

PooLP:
The string must contain the technical accountName (ex: i: 0#.f|membership|george.dumont@tenant.onmicrosoft.com), but it may not be the best way, if tou have another idea, say me ^^

FranckyC:
Hi @PooLP. So I took the time to review your PR and here are my comments.

- It does not work in local workbench. Maybe you should add your mock data in the search service and add refinement results with relevant claims values.
- A documentation could be nice (`search-refiners.md`). Because the Persona refiner template seems to only work with specific crawled properties containing claims values (like 'i:0#.f|membership|gradya@aequosdev.onmicrosoft.com'), maybe you should list them in the documentation to help people to use it. For instance:
	- People:Manager (only available with People result source)
	- People:AccountName (only available with People result source)
	- Any `ows_q_USER_<column_name>` property (like ows_q_USER_Author)
	- etc.

- Rename the `SpService` by `UserService`. The current folder name is not relevant.
- Once selected, there is no indication the filter is actually selected. Maybe set the text to bold in this case.
- Modify _"File types"_ and _"Multiple file types"_ resources text to singular instead of plural.
- All `IUserProperties` interface members could bejust replaced by a single `[key:string]: string` instead of listing all available properties.
- In the filter layout, to display the user name instead of the claim value for the selected user, in the `PersonaTemplate.tsx`, you can just grab the user name from the claims raw value and update the `refinementValue` variable before returning it:

```
	let displayName: string = refinementValue.RefinementValue.split('|').length > 1 ? refinementValue.RefinementValue.split('|')[1].trim() : null;
	...
	
	return (
              <PersonaCustom
				...
                onClick={(ev) => {
                  refinementValue.RefinementName = displayName ? displayName : refinementValue.RefinementName;
                  this._onFilterAdded(refinementValue);
                }}
              />
            );
	
```
You can use the same concept for the FileType refiner (using `FileHelper.extensionToLabel(extension)`).

With these changes, we are go to go ;). Thanks for the help Richard.

PooLP:
It's ok for all except selected state for Persona, I would suggest something else later (with "multiple" version)

I encountered a problem with latest version of dev branch, i reverted to pnpjs 1.3.4, there is a bug since the version 1.3.5 with taxonomy. The GUID passed to service didn't work correctly.

I did not have the time to see this.

image

wobba:
I reworked another issue with 1.3.8.. feel free to revert to 1.3.4, and we should possibly log issues with pnpjs as well if there is a bug.

FranckyC:
Hi @PooLP, I finally merged your PR! However, I did some updates about the selection behavior + fixes some typos in the doc. For the selection, I removed the _'Clear'_ button and replaced by a bold font when selected.

FranckyC:
But thanks for this PR, looks really nice though ;)

FranckyC:
@PooLP I just figured out there is a glitch when the refiner value is the system account and doesn't have a proper account name. I don't have time right now to investigate but if you have time to render system account as well (this is a valid refiner after all), this would be appreciated

FranckyC:
@PooLP FYI, I'm working on a refactoring to improve this scenario + performances.

PooLP:
Sorry for delay, i'm in vacation for ~3 week.
I see that you had refactor, it's much better like that. 
For my next participation, I will make / correct renderings, it's more my expertise :D.
See you soon.

PR #41, opened 2019-12-13
Thumbnail height is incorrect

stevebeauge:
Here's a minor update to fix an issue regarding the height in tile views in compact mode.

Before:

image

After:

image


FranckyC:
Thankls @stevebeauge !

PR #59, opened 2020-01-15
[Search Refiners] Refactored the 'Persona' template to improve performances

FranckyC:
- Use a batch request to retreive profiles proeprties instead of making a call for each item.
- Remove the PersonaCustom.tsx since it is not needed anymore.


FranckyC:
@PoolP I've made some updates to your original code mainly to improve performances and show non valid claims values (like 'System Account'). Thank you again for the initial PR ;)

PR #67, opened 2020-01-21
Query modifier

bschlintz:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no
| Related issues?  | n/a

#### What's in this Pull Request?

Add support for a custom query modifier via extensibility library. This capability allows developers to write a query modifier that intercepts the user's search query and optionally modifies it before it is sent to SharePoint to fetch search results. The query modifier may manipulate the query text and/or the query template.

Includes the following:
- Updated extensibility library project and docs to demonstrate custom query modifier capabilities
- Added 'Query Modifier' section to search results web part settings pane that is only visible when a query modifier is deployed. This section shows the name and description of the query modifier along with a toggle to disable it if needed.

Additionally, the following unrelated change is included:
- Added 'currentPage' property to the Handlebars template context so a template may conditionally render certain elements based on page number (e.g. only show secondary results on page 1).


FranckyC:
@bschlintz This is really cool! Works well. Maybe I should refactor the Query Enhancer feature to work with this since it can be redundant/confusing for the user to have multiple solution for basically the same purpose. But definitely like your approach! Also it seems very similar to the new SPFx query suggestion https://docs.microsoft.com/en-us/sharepoint/dev/spfx/extensions/get-started/query-suggestion-tutorial-building.

bschlintz:
@FranckyC thanks for reviewing and merging! I thought the same; perhaps this would be a good mechanism to hook in the NLP service in the example project. 

For our use case, we wanted to have access to the original query text, template and result source ID to make decisions about how to modify the query. Since the template and result source ID aren't available on the search box since they can be set by the verticals or results web parts, we needed to insert ourselves just before the query fired so we could see all the properties from the various sources. We also needed the ability to modify the query template in addition to the query text for some transformations. This model is very similar to the first-party extension, but takes it a little further by exposing more than just the query text for viewing and modifying (we hope to migrate to the first-party extension when all the features we need are there).

kmprasad4u:
We are considering to use Search Results web part as a Content Search Web Part for content roll-up and not specifically for Search Results Page purpose. So it would be used in multiple pages in same site itself.  

Based on requirements, we are in need of modifying the search query for specific web part in a page. We are trying to avoid modifications on the source code of the search web parts to reduce impact for merging future changes in the solutions.

Since the query modifier extension library is deployed across site / tenant and uses a single library component id, the chances of the modifier affecting the other search results web parts are high. Is it possible to have the modifier implemented as a web part connection instead of extension library. With that approach, we could also easily target multiple search results web parts based on needs. 

Another option is to have a property in the search results web part that will allow to key in the component ID of the library.

kmprasad4u:
@FranckyC 

We are considering to use Search Results web part as a Content Search Web Part for content roll-up and not specifically for Search Results Page purpose. So it would be used in multiple pages in same site itself.

Based on requirements, we are in need of modifying the search query for specific web part in a page. We are trying to avoid modifications on the source code of the search web parts to reduce impact for merging future changes in the solutions.

Since the query modifier extension library is deployed across site / tenant and uses a single library component id, the chances of the modifier affecting the other search results web parts are high. Is it possible to have the modifier implemented as a web part connection instead of extension library. With that approach, we could also easily target multiple search results web parts based on needs.

Another option is to have a property in the search results web part that will allow to key in the component ID of the library.

kmprasad4u:
Opened another thread for the question, since this is closed

Issue #57, opened 2020-01-13
Redirecting search queries to PnP Modern Search results page only fires once


colonelclaypoo:
Hello,

I'm not sure if this is an issue but I wanted to make you aware of the following:

Based on Mikael Svensons Ignite presentation I was eager to learn that you can redirect search queries to a PnP Modern Search results page. Has anyone encountered the issue that it only works for the initial query?

So from site collection A I'm entering a query that is passed on to site collection B which hosts the results page. The parameters passed are Home.aspx?k=test&ql=2055, for example.

I have configured the result source web part to use the page environments query parameter k.

![image](https://gxcuf89792.i.lithium.com/t5/image/serverpage/image-id/165011iC8AA0B8FB58BF790/image-size/large?v=1.0&px=999)

The search results web part correctly shows hits for the query. However, I can't get a new query to fire.

When I use the out of the box modern page's search box the url in the address bar changes but it's not passed on to the search results web part:

![image](https://gxcuf89792.i.lithium.com/t5/image/serverpage/image-id/165018iA1E0F33AA6A4D899/image-size/large?v=1.0&px=999)

When I refresh the browser page the new query is passed on and the web part shows the new results.

Is anyone experiencing this behaviour?

I though of putting a PnP Modern Search box onto the page for further queries but the results web part can only be connected to one source it seems.

Thanks.

FranckyC:
Hi @colonelclaypoo, thanks for your feedback. Actually this due to the way the search box is implemented and also  the way SharePoint handles page redirection via its own page router. By default, the search box  "listens" the URL only when the URL hash changes ("#"). Unfortunately it doesn't work for query parameters because there is no page postback when you submit your search.

This is definitely something we should improve to get a nice experience.
@wobba do you know an "SPFx" way (via the page router) to catch the event when a query string parameter is changed?

For the hash I use the builtin listener `hashchange`.

wobba:
Could be that some of Elio's approaches from https://www.eliostruyf.com/handling-navigation-in-a-sharepoint-framework-application-customizer/ work for query param changes.

FranckyC:
Yeah but it will introduce a new component in the solution and again, add more complexity to the whole solution. It is too complicated to resolve only this issue in my opinion. We need to think about an other approach, lighter.

wobba:
@FranckyC there's no event in SPFx for this. I'll ping you offline on this.

colonelclaypoo:
[...]

PR #68, opened 2020-01-22
Search Queries update URL when Dynamic Data is Enabled

bschlintz:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no
| Related issues?  | n/a

#### What's in this Pull Request?

The URL will now update when the search query is changed in the Search Box. This will only happen when the Search Box has dynamic data enabled and uses the URL hash or a query string parameter. The dynamic data configuration is detected when a search executes and the URL history is updated accordingly.


FranckyC:
Thanks @bschlintz . Does this resolve this issue as well https://github.com/microsoft-search/pnp-modern-search/issues/57 you think?

bschlintz:
Unfortunately, I don't think this PR will solve that issue @FranckyC . This PR adds logic to the PnP Search Box web part to update the URL as the search text changes.

FranckyC:
@bschlintz I will process it as soon I have time, Thanks again!

PR #78, opened 2020-01-29
Update templating.md

KEMiCZA:
Might save some time for other people. It was not obvious that we had to prefix pnp- for the webcomponents.


wobba:
@FranckyC can you verify that the prefix is now needed?

FranckyC:
You are right. Actually, you need a prefix but it doesn't have to be `<pnp->` necessarily.

Issue #83, opened 2020-02-01
Persona refiner error if email is not the same as account name


patrikhellgren:
**Version used**
3.9

**Describe the bug**
If the user email address is not the same as the account name in the claim the Persona refiner will return an error and the refiner webpart will crash.

**To Reproduce**
1. Make sure the user has an email address different than the account name
2. Map a people field crawled property to a RefinableString managed property
3. Set the people field to the user with different email address and account name
4. Wait for a crawl
5. Define a Persona refiner for the RefinableString
6. Do a search and try to refine with the Persona refiner and the refiner webpart will go blank and an error will be returned in the console.

**Expected behavior**
The Persona refiner should work just as well with users with account names other than the email address.

**Desktop (please complete the following information):**
 - Browser Chrome, Edge

I will provide a pull request with a fix for this, just wanted to register an issue if others run into the same issue.

FranckyC:
Thanks @patrikhellgren . I didn't test this case ;)

PR #84, opened 2020-02-01
Use claim and account name instead of email in Persona refiner

patrikhellgren:
This is a fix for issue #83


FranckyC:
Thanks @patrikhellgren. Works well ;)

PR #86, opened 2020-02-05
new Handlebar Helper - getAttachments

joaojmendes:
New Handlebar Helper to get Attachments from a string


wobba:
Could you explain a bit more what the function does and what it's used for? Also, your commit seems to mess with git formatting making it hard for git to compare the sources and find the actual changes you  made.

joaojmendes:
Hi Mikael,

I needed to get attachments from list item and created a function to get all attachments from "LinkOfficeChild" and, for each attachment, returns an object with URL and file name.

About git I will check what is happening .

Thank you
João

FranckyC:
Hi @joaojmendes . I'm reviewing PR and I have a couple of remarks:
- The documentation isn't very clear. I had to look at the code to figure out how this is supposed to be used. Maybe you should mention that is an Handlebars block helper intended to be used with the context variables it provides like this (url and filename):

```
{{#getAttachments LinkOfficeChild}}
    <a href="{{url}}">{{fileName}}</href>
{{/getAttachments}}
```

Also you should mention in the doc the helper is intended to be used only with the `LinkOfficeChild` managed property wich is the default managed proeprty for list attachments.

joaojmendes:
Hi Franck,

I updated the Docs with new text.

Thank you very much.

João

FranckyC:
Thanks @joaojmendes !

allafridman:
I am getting the error Error: Missing helper: "getAttachments"

wobba:
@allafridman log a new issue and state your version etc.

PR #95, opened 2020-02-13
Fix bug when selecting an item from the Sort dropdown

joelfmrodrigues:
Fix bug when selecting sortable dropdown items for the first time - it was defaulting to the first item as per the following thread https://github.com/OfficeDev/office-ui-fabric-react/issues/5473 

Added new dropdown option that is hidden and selected by default.

ezgif com-optimize


FranckyC:
Works great thanks @joelfmrodrigues !

PR #108, opened 2020-02-21
Default sortable field and direction

### LINKED ISSUE #107: Default sortable field and direction
Based on user feedback, I would like to propose the following:

1. Default sortable field dropdown value
If a field is used as the default sorting on the results web part, it would be great if the sortable fields dropdown would automatically default to that field and order (assuming the same field is added on both sort configuration options).

2. Default sortable field direction
It would be great if a sortable field could be configured to have a default direction.
As an example, if you currently want to add a sortable field of date/time (for instance Modified), it will be ascending by default, which is unlikely to be the desired result so users have to always invert the direction.

joelfmrodrigues:
Fixes #107 

Setting default sortable field and direction with the same field as set in the default results sort (if same field is the same option on both configurations)
Added new column to set default direction for sortable fields - useful for example when dealing with dates as descending is often more useful

ezgif com-optimize


FranckyC:
Hi @joelfmrodrigues I did some tests and I noticed sometimes the initial sort order fields or sortable fields are not saved in the Web Part after configuration. Quickly it could be due to mutating objects due to a wrong JS assignation between properties `sortList`/`sortFields` in the WP lifecycle. Could you please review variable assignations and use immutable objects when needed? Thanks!

joelfmrodrigues:
Hi,

Sure, I will have a look. Thanks for testing it

joelfmrodrigues:
@FranckyC sorry but I am a bit lost with this
I tried to configure the web part many times and the only case where sort/sortable configurations were not saving was when one of the required fields was not populated (i.e. Direction). Other than that, all the settings were always saved.
Are you able to provide more detail on how to recreate the issue?

FranckyC:
Hi @joelfmrodrigues After redoing some testing, you're rught I'm not able to reproduce the issue... My bad, I apologize for the invested time on this. I was probably not well awake when testing the first time :).

joelfmrodrigues:
@FranckyC that's absolutely fine, I'm sure it took me way less time than having to write the whole solution 😄 what a great project this is!
Thanks for being so quick to provide feedback and process the PRs!

PR #90, opened 2020-02-10
Added possibility to define default selected filters using the url parameter

agtenr:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| Related issues?  | /

#### What's in this Pull Request?

This PR wast created based on [this](https://github.com/SharePoint/sp-dev-solutions/issues/323) question.

As a first set-up i added the possibility to add pre-selected filters in the url for the Checkbox and Date filter types. When the refiner web part is loaded, the url parameters are checked, processed, evaluated and sent to the search service. You can define the default refiners using the `filters` url parameter. This parameter should contain a JSON string with the different filters.

e.g. `filters=[{"filterName":"Department","filterValues":["HR","Legal"],"filterOperator":"or"}]`

 - filterName: the name of the refiner (also used in the configuration of the web part)
 - filterValues: Can be a string or an array of strings. 
 - filterOperator: is optional. If not provided, the operator from the web part configuration will be taken into account

##### filterValue
The filter value for the date range should be in the following format: 'range(date 1,date 2)':
 - Date 1: Should be an ISO formatted UTC date OR min
 - Date 2: Should be an ISO formatted UTC date OR max

The filter value for a the date fixed interval should have one of the following values:
 - yesterday
 - weekAgo
 - monthAgo
 - threeMonthsAgo
 - yearAgo
 - olderThanYear

#### Next steps
 - Add support for other refiner templates (file type, persona, ...)
 - Update documentation

Please let me know what you guys think. Thanks

Issue #48, opened 2020-01-03
Support for deep linking refiners (configuring refiners from the query string)


rudfoss:
**Is your feature request related to a problem? Please describe.**
I can't seem to find any documentation on if it's possible to control which refiners are active based on url query parameters (deep linking). Is this possible and if so how would I go about using it?

**Describe the solution you'd like**
Support for deep linking into specific refiners as well as some documentation on how to use it.

**Describe alternatives you've considered**
I've considered building a custom web part to handle this or simply hard coding a search into the results web part.

FranckyC:
Hi @rudfoss. Currently this is not possible. Thank you for the feedback, we will think how to handle this scenario correctly for next releases.

rudfoss:
@FranckyC Thanks for the reply. Any chance of getting a timeline for this feature? Would it be hard to implement as a PR, do you think? I haven't looked at the code much yet.

FranckyC:
It could be depending what we want to support. For now, no timeline for this feature. If you think about a specific schema, please let us know before starting coding and then you could submit a PR once validated.

FranckyC:
Fixed by #116

PR #116, opened 2020-02-26
[search-parts] Refinement filters deep links

FranckyC:
Added the ability to created deep links for refinement filters using URL query string parameter `filters`:

- Should work with/without search filters WP.
- If search filters WP present, should show selected values automatically.
- Should handle the initial selection vs user selection scenarios.
- Should handle FQL and raw text conditions.
- Should work with multiple conditions for a single managed property.
- Should respect the multi/single value filter template setting. If multivalues are provided to a single configured template, only the first value is used.
- Should fail nicely if a wrong JSON structure is provided in the URL.
- Should handle all refiner templates.
- Should expand when a value is selected regardless the 'Show expanded' value configured for the refiner.
- Should support multiple refiner managed properties in the URL.
- Should provide a documentation on how to use the URL parameters.
- Should work seamlessly with query keywords query string parameters.

Related #90 #48.


arnieraju:
@FranckyC I have just started testing out your modern search web parts and they are incredible.  Thank you so much for sharing!

After some initial testing my first question came up about, use of the search queries and refiners, deep linking and browser history. 

It seems my question came up at the same time you were merging this PR, which handles the deep linking side of things which is awesome, I can't wait to try this out.

This leaves browser history side of things.  What I am looking for here is something along the lines of the following:

1. User searches for "Hello world", address bar is updated to https://contoso.sharepoint.com/SitePages/Search.aspx?q=Hello%20World and this is added to browser history
2. User refines by Date Created, address bar is updated to reflect refinement choices and new url is added to browser history.

I had a very quick scan of the code and could not see provision for this in the code, that being said I may have missed it :)  

1. Is there any work being done, or are there any plans for implementing updates to the browser history as users interact with the search query web part or refiners?  
2. If not I'd love to help contribute in some way.

FranckyC:
Hi @arnieraju this would be a great addition for sure. I did not think about it when implementing this feature. For now, the URL filters are only read at page load. If we want to use the browser history like you described, it shouldn't interfer with the dynamic data mechanism used to exchange refiners/refinement results between the search filters WP and the search results WP. It means the logic should only care about the 'initial' and 'backward' scenarios. Feel free to start something if you want on your side if you have an idea how to handle this. I'll be happy to review/complete your PR.

PR #125, opened 2020-03-05
Filter Checkbox refiner options by name

joelfmrodrigues:
Added text box to allow filtering list of refiner options by name. This is very useful when the list of refiner options is large (i.e. 100 options).
The text box is only visible when there are more than 10 refiner options.

image

When checkbox group is set to multi-select, filter ignores selected items and those are always visible

image

### DIFF
--- a/search-parts/src/webparts/searchRefiners/ISearchRefinersWebPartProps.ts
+++ b/search-parts/src/webparts/searchRefiners/ISearchRefinersWebPartProps.ts
@@ -5,6 +5,7 @@ export interface ISearchRefinersWebPartProps {
   webPartTitle: string;
   showBlank: boolean;
   refinersConfiguration: IRefinerConfiguration[];
+  showFilterBoxForRefinerValues: boolean;
   searchResultsDataSourceReference: string;
   selectedLayout: RefinersLayoutOption;
 }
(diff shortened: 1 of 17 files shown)


wobba:
Is there a web part setting to toggle filter on/off, and update docs as well to include the new filter. And great addition 👍

joelfmrodrigues:
Just a bit flooded with work at the moment and had to swap some priorities...will look into this in a few days :)

joelfmrodrigues:
@wobba just updated the PR. I noticed that I ended up making multiple commits so let me know if this is a problem and I will squash them

FranckyC:
Hi @joelfmrodrigues, after a quick test, it seems there are some issues with the toggle state:

filters

joelfmrodrigues:
Thanks @FranckyC I haven't noticed that before.
Will have a look

joelfmrodrigues:
@FranckyC just pushed a new commit that should resolve this. I was not passing the correct property instance to TemplateRenderer from Vertical and LinkPanel.
I must been sleeping when tested that as it was behaving opposite to the expected :)

wobba:
@joelfmrodrigues could you please update and resolve with current code?

joelfmrodrigues:
@wobba I thought this PR was abandoned as Frank merged a PR from another user to add similar functionality.
@FranckyC please shout if I'm wrong, but I think that the other PR was only missing the option to keep selection always visible while filtering multi selects. I am still planning to implement that at some point if not done already, just delayed by other work

wobba:
You are right 🙂

PR #131, opened 2020-03-16
[search-parts] [searchVerticals]: Removing white space if no icon selected  

NizarGrindi:
Removing placeholder (whitespace) if the icon property is empty or wrong for the search verticals

### Before
Before

### After
After


wobba:
Would
```
if (vertical.iconName)
```

be better as it also checks null. Or even
```
if (vertical.iconName && vertical.iconName.trim() !== "") {

}
```

wobba:
Thank you @NizarGrindi :)

PR #144, opened 2020-03-24
Added support for faster development with a separate webpack dev server.

s-KaiNet:
This PR doesn't add any new functionality, but it tries to improve the development speed.  
   
I tried to improve the performance of the default SPFx builds using different ways.   
The latest and fastest approach is the tool [spfx-fast-serve](https://github.com/s-KaiNet/spfx-fast-serve).    

I used `pnp-modern-search` solution to test performance gain and the test demonstrated significant performance improvements. Since I have the code, which speeds up serve for `pnp-modern-search` locally, I decided to push a PR, so that you can try `fast-serve` and maybe even merge the PR.   

Basically you just need to run `npm install` to restore additional dependencies and then run `npm run serve`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@ obj
 
 # Styles Generated Code
 *.scss.ts
+*.scss.d.ts
 
 # compiled output
 dist/
--- a/search-parts/gulpfile.js
+++ b/search-parts/gulpfile.js
@@ -7,6 +7,7 @@ const build = require('@microsoft/sp-build-web');
 const log = require('@microsoft/gulp-core-build').log;
 const bundleAnalyzer = require('webpack-bundle-analyzer');
 const colors = require("colors");
+const fs = require("fs");
 let buildConfig = {
     parallel: os.cpus().length - 1
 };
@@ -21,6 +22,8 @@ const envCheck = build.subTask('environmentCheck', (gulp, config, done) => {
     }
     build.configureWebpack.mergeConfig({
         additionalConfiguration: (generatedConfiguration) => {
+
+            fs.writeFileSync("./temp/_webpack_config.json", JSON.stringify(generatedConfiguration, null, 2));
 
             /********************************************************************************************
              * Adds an alias for handlebars in order to avoid errors while gulping the project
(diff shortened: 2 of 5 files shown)


wobba:
Thank you and we will take a look at this and evaluate.

FranckyC:
Thanks for this PR. It is truly amazing how things are fast now! You made my day ;). However I noticed the `SearchResultsTemplate.scss` file is not compiled with this config, breaking the 'Simple List' layout. Maybe related to the used loaders?

image

s-KaiNet:
Yes, I see the issue... `css-loader` treats regular `.scss` as a module, thus breaks all css classes. We need css modules only for `*.module.scss`. I will make a fix.

FranckyC:
Hi @s-KaiNet , this seems to resolve the problem, thanks. Quick question: how I can use the `--nobrowser` flag like the regular `gulp serve --nobrowser`?

s-KaiNet:
Sure, you can just set `open: false` in `webpack.js` file for [devServer](https://webpack.js.org/configuration/dev-server/#devserveropen) configuration.

FranckyC:
Got it! However, I'm not able to debug using VSCode Chrome debugguer. The source maps are correctly loaded because I can debug with the browser but VSCode doesn't recognize my breakpoints :(. My `launch.json` config:

```
 {
      "name": "Hosted workbench",
      "type": "chrome",
      "request": "launch",
      "url": "https://<mysite>/_layouts/workbench.aspx",
      "webRoot": "${workspaceRoot}",
      "sourceMaps": true,
      "sourceMapPathOverrides": {
        "webpack:///.././src/*": "${webRoot}/src/*",
        "webpack:///../../../src/*": "${webRoot}/src/*",
        "webpack:///../../../../src/*": "${webRoot}/src/*",
        "webpack:///../../../../../../src/*": "${webRoot}/src/*",
        "webpack:///../../../../../../../src/*": "${webRoot}/src/*",
        "webpack:///../../../../../../../../src/*": "${webRoot}/src/*",
        "webpack:///../../../../../../../../../src/*": "${webRoot}/src/*"
      },
      "runtimeArgs": [
        "--remote-debugging-port=9222",
        "-incognito"
      ]
    }
```

Any idea?

s-KaiNet:
Good catch!   
I found out, that vscode chrome debugger uses source maps loaded from disk, i.e. `c://<project path>/dist/search-wp.js.map`. However, with custom webpack build all maps are in memory and `dist` folder contains maps created by the `gulp bundle` process (which are not the same as the ones created by custom webpack.js build). The trick was just to delete all files inside `dist` before running custom webpack build. To test it, you just need to restore deps and run `npm run serve` and vscode debugging.   

Additionally, I added a small fix (also with `del` module) for the issue, reported in the repo by another developer.

FranckyC:
Works well now. Thank you for this addition, it is really nice! You will make SPFx developers happy ;)

wobba:
Nice :D I had to run gulp clean before running, but this helps a lot!

Would be cool if `gulp serve` could kick off this one instead. Did we add documentation on using `npm run serve`?

s-KaiNet:
In theory, it's possible to integrate these things to `gulp serve`, however it requires a lot of SPFx build pipeline tasks to be re-architected and re-implemented. So in practice, I'm not sure it's possible today, but maybe in the long term yes. SPFx grows quickly, thus performance problems with `serve` will be even more relevant in the future when a codebase grows.    

JFYI I found some other things, which need to be fixed regarding `fast-serve`. I will push a separate PR later today or tomorrow.

FranckyC:
Like @wobba pointed out, a nice thing would be also to wrte a quick paragraph in the doc about the `npm run serve command`. Feel free to add a link to your blog posts as well.

Issue #145, opened 2020-03-25
Modern Search: Refinement value filter


stijnbrouwers:
When having a lot of values available for a refinement, it can be cumbersome to navigate through all of them to find the desired value to apply. 

It would be nice if for each refinement we can have an option 'show value filter' that, when checked, it shows a textfield. The textfield can be used to enter a string that will be used to match against the refinement filter values. Only the values will be shown that contain the value filter string.

![image](https://user-images.githubusercontent.com/9864443/77504731-75797a80-6e61-11ea-9f6f-3cb756e1ece5.png)
The above image shows a filtering on 'cust' which will only show refinement values that contain the string

FranckyC:
Resolved by #146

PR #146, opened 2020-03-25
Add refinement value filters to each template

stijnbrouwers:
**What's in this Pull Request?**
Refinement value filters added to reduce the number of refinement values shown to the user

**Related issue**
Implements #145

### DIFF
--- a/search-parts/src/models/IRefinerConfiguration.ts
+++ b/search-parts/src/models/IRefinerConfiguration.ts
@@ -34,6 +34,11 @@ interface IRefinerConfiguration {
      * Allow refiners to be expanded by default
      */
     showExpanded: boolean;
+
+    /** 
+     * Show filter textbox to search inside the refiner values
+     */
+    showValueFilter: boolean;
 }
 
 export default IRefinerConfiguration;
--- a/search-parts/src/webparts/searchRefiners/SearchRefinersWebPart.ts
+++ b/search-parts/src/webparts/searchRefiners/SearchRefinersWebPart.ts
@@ -315,6 +315,11 @@ export default class SearchRefinersWebPart extends BaseClientSideWebPart<ISearch
             id: 'showExpanded',
             title: strings.Refiners.ShowExpanded,
             type: CustomCollectionFieldType.boolean
+          },
+          {
+            id: 'showValueFilter',
+            title: strings.Refiners.showValueFilter,
+            type: CustomCollectionFieldType.boolean
           }
         ]
       }),
@@ -456,7 +461,8 @@ export default class SearchRefinersWebPart extends BaseClientSideWebPart<ISearch
           template: RefinerTemplateOption.CheckBox,
           refinerSortType: RefinersSortOption.Default,
           refinerSortDirection: RefinersSortDirection.Ascending,
-          showExpanded: false
+          showExpanded: false,
+          showValueFilter: false
         },
         {
           refinerName: "Size",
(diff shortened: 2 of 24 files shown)


FranckyC:
Hmm looks very similar to #125  already submitted by @joelfmrodrigues.

joelfmrodrigues:
At the moment I am only adding this to Checkbox, but should now be easy to replicate to the other templates

stijnbrouwers:
Sorry guys, I first submitted it to sp-dev-solutions before I noticed it was moved. I forgot to check the open issues here...

The solutions look very much alike in implementation as well :-)
Difference I see is:
- @joelfmrodrigues keeps selected values visible when they don't match the filter (mine does not)
- I implemented a clear button to quickly remove the filter value
- I allow to enable/disable the filter per refiner, @joelfmrodrigues enables it for all supported filters at once
- I don't save the filtered values in the state, only the filter value which I apply during the render method
- I added the change to Persona and file type template as well as checkbox

Let's cherry pick the preferred differences above and determine from there what PR to keep? 

I am really sorry for the inconvenience...

FranckyC:
@stijnbrouwers It looks like really nice actually. A detail, can you just apply a `10px` padding on the textfield to get left aligned with check boxes?

@joelfmrodrigues This solution uses a per filter setting which I think is more flexible than an unique toggle for all. I will go over this PR instead of this one #125 .

Also, I can't merge the two unfortunately. @joelfmrodrigues feel free to enhance with your ideas after merge. Hope you understand. Team work here ;)

Anyway, nice work guys! Reallky appreciated.

joelfmrodrigues:
That's absolutely fine from me Frank, this is a more configurable approach and already has support for more templates 👍
I still think that always showing selected items is a valuable behaviour as otherwise it's harder for end users to know what is currently selected.

stijnbrouwers:
@joelfmrodrigues ,

I played with the idea as well...
On one hand, it's more clear because it's visual to the end user what was selected. On the other hand, the filter is to limit the number of values shown without the need to scroll through them. If you have 10 values selected and the filter reduces the number of values shown to 10. You still have 20 values visible, which may be too much. 
Both have their merits, in my opinion, it's just a matter of making a choice.

I also played with the idea of showing selected values below or above the filtered values with some kind of visual feedback (selected values & filtered values or a horizontal line to separate the two). But I came back from that idea because this would make the behavior different from when no filter is applied. 

Because I haven't decided yet what is the best scenario, I thought to just submit it like this for now and come back to it later.

I would appreciate any feedback on this though :-)!

stijnbrouwers:
@FranckyC 
I added the padding left to the file type and checkbox templates.
I didn't do it for the persona template to keep it aligned with the values here as well.
Check following screenshot of the result:
image

joelfmrodrigues:
@stijnbrouwers regarding displaying the selection, I agree to some extent. I personally think that displaying 20 items (10 selected + 10 filtered) still gives more value to the user. Seems fairly easy to visually "filter out" the items already selected and focus only on the remaining ones. I think it would be harder to mentally manage a list of selected items if we have 10 options already selected.
And in my case it was a client requirement :) So I will need it :)

FranckyC:
@stijnbrouwers @joelfmrodrigues I will go with this PR for now. We could still enhance the behavior in the next release based on feedback.

FranckyC:
@stijnbrouwers It seems your CSS is not applied on the search box:

image

image

Just remove the quotes.

stijnbrouwers:
@FranckyC Fixed (sorry for the weird commit message, I mistyped my git command...)

FranckyC:
Thanks!

PR #155, opened 2020-03-28
[search-parts] Refactoring

FranckyC:
- Replaced the pagination Web Part by a web component. You can still decide where to put the component in the template + additional property pane options.
- Removed the search navigation Web Part. Replaced by the search verticals Web Part.
- Added an `instanceId` information in the Handlebars context for templates to be able to scope CSS styles per Web part instance.
- Updated the documentation.


stevebeauge:
This introduces major breaking changes (existing web templates has to be reworked) and pagination webpart disappeared.

I won't discuss the why, but should highlight in bold blinking red this fact.

(unless I missed something)

FranckyC:
I'll add to the doc + release

stevebeauge:
I know this won't be trivial, but maybe you can introduce a "deprecated" state of now-old webparts, to avoid reconfigure in unique pass all usages of the component (especially if deployed at tenant level).

wobba:
If we have breaking changes we need to highlight them, or change versioning when doing so.

FranckyC:
In terms of code refactoring this is really a challenge to integrate deprecated stuff with new things honestly. The code base is no so flexible. I know this sounds a pretty aggressive approach to wipe them out so quickly but I have no choice if I want things to move forward. Because it is open source and there is no official support, I took the liberty to do that. You can stay on 3.10.0 for now. But you're right, I should add a documentation on how to make the transition (at least for the pagination WP). Here it is: https://microsoft-search.github.io/pnp-modern-search/search-parts/search-pagination/. The migration shouldn't be a big deal if you used a centralized file for custom templates. I also updated the release note to include breaking changes.

wobba:
We've made notes before between versions to take special care. And going to v4 seems to big a leap indeed. We can start informing one version up front in the future probably.

Georg-Trixl:
About removed **navigation web part** 
Okay you guys did a nice job at all. With the navigation web part it was possible to recreate in modern looking pages something like _verticals_ . On each page it was possible to create a total different Result style (Card-Style, List-Style, Person-Cards).

But now the **navigation web part** is removed - how to build a all in one dynamic template that shows Persona-Cards on the People vertical, Document-Cards on Videos vertical and List-Style on Recently changed items?

wobba:
@Georg-Trixl You can use the result type matching - set an expression on what to match for, and use the template for this. I have a sample pnp template using this if you are interested.

Georg-Trixl:
@wobba yes i would be very glad, if you could share your sample pnp template.

wobba:
@Georg-Trixl  Try this one which creates a page named Search.aspx https://1drv.ms/u/s!AsQPRl4COMyeoqIHK_uoGsw22Mm8iw?e=JhmtVT

roqz:
What about search custom renderer implementations and pagination? did I miss something or is there no way to enable pagination after the update?

Georg-Trixl:
I added an extra search results webpart, to my results page, then i switched in the new, webpart to similar results-layout and looked into the template code - there was also pagination code - added that to my custom renderer, did the job.

roqz:
can you show an example? the custom renderer examples do not have dependencies on any pnp-modern-search package, so where do you get the context information containing the pagination data from?

Georg-Trixl:
okidoki, take a look into the current display templates of this source code, search for  <pnp-pagination, for example here
[https://github.com/microsoft-search/pnp-modern-search/blob/master/search-parts/src/templates/layouts/simple-list.html](https://github.com/microsoft-search/pnp-modern-search/blob/master/search-parts/src/templates/layouts/simple-list.html) there you find your answer

roqz:
that is a custom handlebars template, isn't it?. documentation also shows how you can use pagination now with custom templates: https://microsoft-search.github.io/pnp-modern-search/search-parts/search-pagination/
what i want to know is how to use pagination if you use search custom renderer implementations (https://microsoft-search.github.io/pnp-modern-search/search-custom-renderer/getting-started/, https://github.com/microsoft-search/pnp-modern-search/tree/master/search-custom-renderer)

wobba:
@FranckyC Seems @roqz has a point here. If you don't use handlebars, how do you now paginate?

roqz:
any news on this topic @FranckyC  @wobba  ?

wobba:
@roqz You can try to output the pagination markup in your component - should work as it's a web component.

roqz:
That does work as a workaround. But as far as i can see the functionality is limited because the webpart settings (regarding pagination) are not available to the custom renderer. That means the only information that is available is the "PaginationInformation" out of the results object of the custom event "ISearchEvent".

wobba:
We're working how best to address it, so for now stay on 3.10.

stevebeauge:
Any update ? 

Or some guidance to implement the workaround (in the documentation of the custom renderer)? 

We lack pagination in custom render to nearly three months...

wobba:
Custom renderer the way we have it will go away for v4, so unless someone wants to code up something to see what needs to be passed along, you're staying on an old version.

PR #156, opened 2020-03-29
[search-parts] template : added missing pagination configuration

NizarGrindi:
After the replacing of de the pagination web part by a web component, the pagination configuration was forgotten in the Custom layout (default.html)


wobba:
Thank you :)

PR #172, opened 2020-04-03
[search-parts] Sanitized HTML in templates + CSS prefixing

FranckyC:
- Added sanitization with DOMPurify for every dangerouslySetInnerHTML
- Added CSS prefixing with Web Part instance ID for custom CSS in templates. Support regular CSS rules + Media queries.

### DIFF
--- a/docs/search-parts/templating.md
+++ b/docs/search-parts/templating.md
@@ -145,6 +145,14 @@ If provided layouts don't meet your requirements, you can modifiy them or start 
 
 ![Edit Template](../images/edit_template.png)
 
+### Styling
+
+You can write your own CSS styles inside templates. However, all CSS rules (including `@media` rules) will be prefixed automatically by an unique ID according to the follwoing pattern (**pnp-modern-search-template_\<Web Part instance ID\>**) to make sure styles are isolated fronm other Web Parts on the page.
+
+#### Sanitized HTML
+
+The HTML is sanitized automatically preventing XSS attacks using [`DOMPurify`](https://github.com/cure53/DOMPurify). It means for instance, you cannot add your own `<script>` tags or inline JavaScript.
+
 ### Handlebars helpers
 
 The following custom helpers are available in addition to the [handlebars-helpers](https://github.com/helpers/handlebars-helpers):
--- a/docs/search-parts/templating.md
+++ b/docs/search-parts/templating.md
@@ -147,7 +147,7 @@ If provided layouts don't meet your requirements, you can modifiy them or start 
 
 ### Styling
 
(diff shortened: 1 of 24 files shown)


FranckyC:
@wobba If you could review this one to make sure we are aligned.

docs/search-parts/templating.md
```diff
@@ -145,6 +145,14 @@ If provided layouts don't meet your requirements, you can modifiy them or start
 
 ![Edit Template](../images/edit_template.png)
 
+### Styling
+
+You can write your own CSS styles inside templates. However, all CSS rules (including `@media` rules) will be prefixed automatically by an unique ID according to the follwoing pattern (**pnp-modern-search-template_\<Web Part instance ID\>**) to make sure styles are isolated fronm other Web Parts on the page.
```
wobba, line 150: Spelling error "fronm"

search-parts/src/components/BaseWebComponent.ts
```diff
@@ -34,11 +34,17 @@ export abstract class BaseWebComponent extends HTMLElement {
...
-                    props[camelCase(attr)] = (value === 'true');
-                } else {
-                    props[camelCase(attr)] = value;
+                // Resolve 'data-*' attribute name
+                const matches = attr.match(/data-(.+)/);
+                if (matches && matches.length === 2) {
+                    attr = matches[1];
+
+                    // Booleans
+                    if (/^(true|false)$/.test(value)) {
```
wobba, line 43: Should this regex be case insensitive?

wobba:
Looks good, but I will do some more tests. Will the ''data" prefix be breaking? If so we should consider how to version it, or if we can auto handle upgrading.

search-parts/src/webparts/searchResults/components/Layouts/SearchResultsTemplate.tsx
```diff
@@ -41,7 +62,55 @@ export default class SearchResultsTemplate extends React.Component<ISearchResult
         let templateContent = props.templateContent;
 
         // Process the Handlebars template
-        const template = await this.props.templateService.processTemplate(props.templateContext, templateContent);
+        let template = await this.props.templateService.processTemplate(props.templateContext, templateContent);
+
+        if (template) {
+
+            // Sanitize the template HTML
+            template = template ? this._domPurify.sanitize(`${template}`) : template;
```
wobba, line 70: Do you need inline if when you have if check on line above?

FranckyC: Probably not ;)

search-parts/src/components/BaseWebComponent.ts
FranckyC: Right this could be /i

FranckyC:
If people used custom web components, yes it will break. I can remove the check of `data-` attributes if too critical.

wobba:
For custom components it should be ok. What if someone used the oob components manually in a template? Maybe we need to mark some functionality as preview to iron out issues when weintroduce it, and still keep versioning simple :)

PR #173, opened 2020-04-04
Default search vertical

LeemansBE:
Hi all,

I've updated the search vertical webpart to give users the possibility to define which search vertical they want to have selected by default. This works depending on a querystring parameter. The value of a given querystring parameter key (configurable in the webpart) will be matched with the current configured search verticals (by property 'tabName'). If a match is found the search vertical will be selected by default. If no match is found the first search vertical will be selected.

image
image

If you've got any additional questions/remarks feel free to ask.

Thx!

Regards,
Bart

### DIFF
--- a/search-parts/package-lock.json
+++ b/search-parts/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "pnp-modern-search",
-    "version": "3.10.0",
+    "version": "3.11.0",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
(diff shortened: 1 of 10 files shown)


wobba:
Would be good if you could take an auto translate stab on the strings. Better that English.

LeemansBE:
Hi @wobba, I've updated the translations. Let's hope google translate did the job properly. :smile:

wobba:
Thank you and someone can always correct it later :)

PR #187, opened 2020-04-15
Default search vertical decode

LeemansBE:
Hi all,

Small update regarding the default query string parameter value.
It should be decoded hence we fetch it from the URL.

Regards,
Bart


wobba:
Thank you.

PR #191, opened 2020-04-16
Improved is file check for determining iconExt

engineererr:
Folders can have dots in their file name as well.
Therefore the check "is folder" should be performed before the "has dots in the filename" (thus is a file) check.


wobba:
Thank you :)

PR #208, opened 2020-04-23
Improve preview handling

wobba:
Improve which URL is returned on getUrl


MakingThingsWorkToronto:
Quick note on the BaseTemplateService > registerTemplateServices change. If ISearchResult item is a List Item it has no file type so there is a null reference on the item.FileType.toLocaleUpperCase() ... need to do a null check here:

image

wobba:
Thank @MakingThingsWorkToronto - fixed :)

PR #209, opened 2020-04-24
Updating React Search Refiners links

KoenZomers:
Currently goes to several redirects/has moved referrers before getting lost :)


wobba:
Thank you 🙂

Issue #214, opened 2020-04-26
Search Results web part: Sort Direction is not reset when select another column from Sort drop-down.


gunner13:
**Version used**
3.11.1

**Describe the bug**
When we select another column in Sort drop-down after having one selected earlier, Sort Direction shows the default value of the new column but results are actually sorted in the old direction.

**To Reproduce**
1. Add a Search Results web part to a page, put an asterisk (*) in the Search query keywords box (just to ensure we have multiple results). 
2. Add multiple managed properties (called column1, column2 from now) to the Sortable Properties list (A->Z for all).
3. Fill in a keyword which returns multiple records, you will see the result are sorted by the column1 and its default direction correctly, then change the direction to Z-A, results are still sorted as expected.
4. Select column2 in the Sort drop-down, you will see it showing A-Z now (which is expected) but results are actually sorted Z-A

**Expected behavior**
Sort Direction should be reset to the default value when we change Sort Column in the dropdown and results should be actually sorted by the default direction.

**Desktop (please complete the following information):**
 - Browser: Chrome (I did not test on other browsers)

PR #216, opened 2020-04-26
Fix for the issue sort direction is wrong when change sort column #214

gunner13:
My commits were based on https://github.com/microsoft-search/pnp-modern-search/pull/183
Fix for this issue https://github.com/microsoft-search/pnp-modern-search/issues/214


wobba:
Thank you :)

PR #239, opened 2020-05-05
Fixed live Persona Web Component

malcom1120:
Fixed mixed case of disableHover property to all lowercase.
Fixed Handlebar Partial registration for pnp-live-persona.

I know the live persona is not supported, but since the code was in the package, I wanted to fix it to work.  This is my first pull request so let me know if I did something wrong.


wobba:
Thank you :)

Issue #237, opened 2020-05-05
'SortList' is not send on API calls towards SharePoint Search.


LeemansBE:
**Version used**
3.12.0

**Describe the bug**
'SortList' is not send on API calls towards SharePoint Search.
The API returns an error:
![image](https://user-images.githubusercontent.com/33486494/81056168-5a6f4100-8eca-11ea-9cf3-99c4fe6ecbbf.png)

**To Reproduce**
1. In the workbench add the 'Search Box' and 'Search Results' webpart
2. Configure the 'Search Results' webpart and connect it to the 'Search Box'
![image](https://user-images.githubusercontent.com/33486494/81056071-314eb080-8eca-11ea-8fa7-8f15ac777f30.png)
3. Open up developer tools and filter the network tab for calls containing 'postQuery'
4. Refresh te page. Check the 'postQuery' request payload and notice the 'SortList' property
![image](https://user-images.githubusercontent.com/33486494/81056373-bd60d800-8eca-11ea-8079-d82501bc9b17.png)
5. Fill in a value in the 'Search Box' webpart and hit enter. This will perform a new 'postQuery' call.
Now the postQuery responds with status '400'. The response looks like 
`{
  "odata.error": {
    "code": "-1, Microsoft.Data.OData.ODataException",
    "message": {
      "lang": "en-US",
      "value": "A null value was found for the property named 'Direction', which has the expected type 'Edm.Int32[Nullable=False]'. The expected type 'Edm.Int32[Nullable=False]' does not allow null values."
    }
  }
}`
Cause:
The request payload does not contain configuration for the 'SortList' property
![image](https://user-images.githubusercontent.com/33486494/81056732-4ed04a00-8ecb-11ea-9a29-684969cc2930.png)

**Additional context**
This problem also occurs when switching between search verticals in the 'Search Vertical' webpart. Possibly the same root cause?

luismanez:
As far as I've seen in a quick debugging, the issue is in file 
_webparts\searchResults\components\SearchResultsContainer\SearchResultsContainer.tsx_
method ___getDefaultSortingValues__ line 712:

```ts
if (this.props.sortList.length > 0
            && this.props.sortableFields.length > 0
            && this.props.sortList[0].sortField === this.props.sortableFields[0].sortField) {
```
__this.props.sortableFields.length__ is empty, so is returning _sortField_ and _sortDirection_ as _null_

[...]

PR #240, opened 2020-05-06
sortlist bugfix

LeemansBE:
Hi all,

This PR is regarding issue https://github.com/microsoft-search/pnp-modern-search/issues/237

The values for _defaultSortingValues.sortField and _defaultSortingValues.sortDirection are only filled up when you have configuration for 'sort order' and 'sortable fields'. When you only got configuration for 'sort order' or no configuration at all the 'sortList' property for the 'Search Service' is still updated with the _defaultSortingValues object. This results in the search query being malformed for the 'sortList' propety.
image

Added the extra check so that the SortList property isn't updated when you don't have 'sortable fields' configured.

Could you take a look at this?

Thx!

Regards,
Bart


wobba:
I did a PR as well at #243 - I tested scenarios where you add/remove/change the sortlist and default settings to ensure there's an update happening for all cases. Does your simpler one work for edits scenarios?

Issue #238, opened 2020-05-05
Bug in Search Results Webpart


Georg-Trixl:
**Version used**
Version 3.12.0 - downloaded today (with the fixed version number)

**Describe the bug**
Error making HttpClient request in queryable [400] Bad Request ::> {"odata.error":{"code":"-1, Microsoft.Data.OData.ODataException","message":{"lang":"en-US","value":"A null value was found for the property named 'Direction', which has the expected type 'Edm.Int32[Nullable=False]'. The expected type 'Edm.Int32[Nullable=False]' does not allow null values."}}}

**To Reproduce**
I have used the workbench and added only the Search Results Webpart.
As Search query keywords I used test
On second page of the configuration no change.
On third panel initially the webpart still renders okay, but if i switch to People layout, the error t comes up.
Below the pictures i describe an other way to get to the error.
![Bug-V3 12-01](https://user-images.githubusercontent.com/62948158/81059557-eb491b00-8ed0-11ea-89ce-3b03a9faf714.PNG)
![Bug-V3 12-02](https://user-images.githubusercontent.com/62948158/81059571-eedca200-8ed0-11ea-87db-d69270e629d8.PNG)
![Bug-V3 12-03](https://user-images.githubusercontent.com/62948158/81059576-f13efc00-8ed0-11ea-8aab-3d352b8e8a06.PNG)

**To Reproduce**
Simple add the _Search Results Webpart_ to the workbench.
As **Search query keywords** use **test**
On second page of configuration panel enter as _Result source_ b09a7990-05ea-4af9-81ef-edfab16c4e31 ( Local People Results )
The error shows immediately up. If I save the configuration and reload the page, then it is rendered without of a error.

**Information of Workbench _Web Part Data - Modern Pages_**

[...]

Issue #234, opened 2020-05-05
Issue while configuring Search result webpart


Narasireddy447:
Have downloaded to solution files and ran npm install command and i got all the node_modules. after that i have build the solution successfully,
i started workbench and while configuring the search result web part, getting following error
"Error making HttpClient request in queryable [400] ::> {"odata.error":{"code":"-1, Microsoft.Data.OData.ODataException","message":{"lang":"en-US","value":"A null value was found for the property named 'Direction', which has the expected type 'Edm.Int32[Nullable=False]'. The expected type 'Edm.Int32[Nullable=False]' does not allow null values."}}}"

please help me if i am missing anything?

Georg-Trixl:
I got the same problem - but i started yesterday already with deeper analysis, to narrow down the problem. A to describe how the development team can reproduce this error. Still working on it - but now I have to go into 2 meetings. Maybe in 2 -3 hours  I have the whole description.

wobba:
For any SPFx project, use the hosted workbench at: https://tenant.sharepoint.com/sites/mysites/_layouts/15/workbench.aspx 

Always a better experience. Let me know if this helps? I have updated the docs as well to reflect this approach in https://github.com/microsoft-search/pnp-modern-search/commit/9b3a458490312cd9eae964b6e80cba3b0873baba

Georg-Trixl:
Okay @wobba got everything to reproduce the error on the simplest way. I hope that helps everyone to deal with it - or maybe there is something wrong on our tenant. 

**Version used**
Version 3.12.0 - downloaded today (with the fixed version number)

**Describe the bug**
Error making HttpClient request in queryable [400] Bad Request ::> {"odata.error":{"code":"-1, Microsoft.Data.OData.ODataException","message":{"lang":"en-US","value":"A null value was found for the property named 'Direction', which has the expected type 'Edm.Int32[Nullable=False]'. The expected type 'Edm.Int32[Nullable=False]' does not allow null values."}}}

**To Reproduce**
I have used the workbench and added only the _Search Results Webpart_. 
As _Search query keywords_ I used **test**
On second page of the configuration no change.
On third panel initially the webpart still renders okay, but if i switch to People layout, the error that @Takeda-scrm-O365-VJE0880 also described comes up.
Below the pictures i describe an other way to get to the error.
[...]

PR #241, opened 2020-05-06
Fix issues regarding the *empty* sort

stevebeauge:
Fix issues regarding the *empty* sort

 #238
 #237
 #234

The idea is consider a `null` sort instead of an empty one

### DIFF
--- a/search-parts/src/webparts/searchResults/components/SearchResultsContainer/SearchResultsContainer.tsx
+++ b/search-parts/src/webparts/searchResults/components/SearchResultsContainer/SearchResultsContainer.tsx
@@ -13,7 +13,7 @@ import { WebPartTitle } from "@pnp/spfx-controls-react/lib/WebPartTitle";
 import SearchResultsTemplate from '../Layouts/SearchResultsTemplate';
 import styles from '../SearchResultsWebPart.module.scss';
 import { SortPanel } from '../SortPanel';
-import { SortDirection } from "@pnp/sp";
+import { SortDirection, Sort } from "@pnp/sp";
 import { ITermData, ITerm } from '@pnp/sp-taxonomy';
 import LocalizationHelper from '../../../../helpers/LocalizationHelper';
 import { Text } from '@microsoft/sp-core-library';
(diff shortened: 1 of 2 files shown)


wobba:
Please look at my PR for the same at #243 as well. I tested scenarios where you add/remove/change the sortlist and default settings to ensure there's an update happening for all cases. Does your solution work for edits scenarios?

stevebeauge:
> 
> 
> Please look at my PR for the same at #243 as well. I tested scenarios where you add/remove/change the sortlist and default settings to ensure there's an update happening for all cases. Does your solution work for edits scenarios?

Yes. Unless I missed some use cases. I checked the edit scenario, and saw sorting as expected. Of course, a careful test session isn't useless

wobba:
@stevebeauge 

1. Have both sortlist and default sort
2. Change default sort to asc/desc
3. Change sortlist to opposite of desc/asc
4. Remove default sort
5. Verify sortlist is used without refresh

stevebeauge:
My test case works with your repro steps

search-parts/src/webparts/searchResults/components/SearchResultsContainer/SearchResultsContainer.tsx
```diff
@@ -705,20 +705,18 @@ export default class SearchResultsContainer extends React.Component<ISearchResul
...
+    private _getDefaultSortingValues() : Sort {
         if (this.props.sortList.length > 0
             && this.props.sortableFields.length > 0
             && this.props.sortList[0].sortField === this.props.sortableFields[0].sortField) {
-            sortField = this.props.sortList[0].sortField;
-            sortDirection = this.props.sortList[0].sortDirection === ISortFieldDirection.Ascending
-                ? SortDirection.Ascending : SortDirection.Descending;
+            return {
+              Property : this.props.sortList[0].sortField,
+              Direction : this.props.sortList[0].sortDirection === ISortFieldDirection.Ascending
```
wobba, line 714: Should be `Direction : this.props.sortableFields[0].sortDirection === ISortFieldDirection.Ascending` I believe.

wobba:
Let me pull your branch and test as well.... less changes than mine in a way.

wobba:
This fails:
1. Add webpart
2. Change edit sort order, set Created to descending. No update on the page.
3. Add a default sort prop, no update/change

I'm inclined to merge my code ;)

PR #242, opened 2020-05-06
226 Enhancement for Secondary Results

malcom1120:
Added support for Custom Results array on Secondary results

Followed the example for RelevantResults and put in results from the CustomResults array if filled for SecondaryResults.

### DIFF
--- a/search-parts/src/services/SearchService/SearchService.ts
+++ b/search-parts/src/services/SearchService/SearchService.ts
@@ -343,7 +343,7 @@ class SearchService implements ISearchService {
                             });
                         }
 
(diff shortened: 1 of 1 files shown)


wobba:
Could both RelevantResults and CustomResults have value.. and should they be concatenated?

malcom1120:
I have not been able to get both to have results at the same time.  Have you seen a scenario (configuration of query rules) where they both get data? I'll test that scenario if I can get it setup.

wobba:
I have no idea, and it's not a scenario I want to advocate anyways :) I'll hold on merging this in light of a quick fix for sorting we have.

malcom1120:
Totally understand waiting for the hotfix.

PR #245, opened 2020-05-06
BaseWebComponent in Extensibility Lib was not considering data-*

luismanez:
The __BaseWebComponent__._resolveAttributes_ method was not considering that now the attributes must start by _data-*_
Custom renderer Layout template updated also with data- (some doc updated too)


wobba:
Thank you :)

PR #246, opened 2020-05-06
Enhance search verticals

malcom1120:
To help with user friendliness:

Added ability that if no defaultVerticalQuerystringParam is configured, try to use the isLink verticals and match to current URL and configured URL


wobba:
I like it and thank you :)

PR #253, opened 2020-05-08
Add New Web Components

malcom1120:
Added Accordion Web Component
- Wraps template HTML in react-collapsible component
- Takes in accordionHeaderText:string, theme:string [Default or Neutral], size:number, startOpen:boolean, roundedCorners:boolean, template:string
- Also made as Handlebar Partial - accordion

Added Popup Web Component
- Divides and wraps template HTML in Office Fabric UI Modal
- Takes in template:string
- Requires HTML marked with data-html-type="popupclick", data-html-type="popupheader" and data-html-type="popupbody" in the template to properly place all the correct HTML for the modal and click
- Also made as Handlebar Partial - popup

Added Lookup List Expander Web Component
- Gives admin ability to expand a search result item from a list/library that contains lookup column for additionial properties from the lookup list
- Wraps the additionial properties returned in the handlebar template
- Takes in listUrl:string [Optional], listId:string, listItemId:number, columnName:string,lookupListFields:string, template:string
- Also made Handlebar Partial - lookuplistexpander

Updated Live Persona Component
Updated Default layout to include examples for Accordion and Popup (can't really give an example for the lookup list expander since it requires a lookup list to be configured)
Updated templating documentation to include new components


MakingThingsWorkToronto:
@malcom1120 - this is amazing! 

I'm upgrading the extensibility layer as we speak. It would be great if you could give it a test early next week. I will defer to @wobba and @FranckyC but some of these may be best in an external library. A goal is to cut down the size of these web parts by not including utilities unless they are common to the core templates.

Adding:
- Plugable support for unlimited extensibility libraries.
- New handlebars helper extensibility.

Let me know if you have any other suggestions!

malcom1120:
@MakingThingsWorkToronto  - Thanks!!

I totally understand about the size and keeping the package smaller.  I added these because I thought they would be highly used by the community.  

I like your idea of unlimited extensible libraries.  It would be awesome to be able to move this to a PnP library that comes as a partner to the webparts with community built web components that you could add-on and use.  

Is it still one library component per tenant catalog? This was my first experience with them and I thought I read somewhere only one library component per tenant catalog.

FranckyC:
Hi @malcom1120, here are my comments:

## Accordion Web Component
- Make react-collapsible a dynamic import since it is not used by default to reduce bundle size
- The react-collapsible theme doesn't really match with the rest IMO and can't be really customized. You could achieve the same using Office UI Fabric grouped list instead of using an external library. Not a big deal but more a general comment.

## Popup Web Component
- Instead of using `data-html-type` attributes to identify popup parts, maybe you should use `<template id="header">` etc. or use [web component slots](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_templates_and_slots).

## Lookup List Expander Web Component
- Could you attach a screenshot with the final result. Not obvious for everyone what it is supposed to do.

## General
- Add [documentation](https://microsoft-search.github.io/pnp-modern-search/search-parts/templating/#web-components) + usage examples (screen should be great). I know existing components don't include this but it could be a start ;).
- No need to create an Handlebars partial every time, the inner content could be processed directly in the web component using `innerHtml`. I used partial before because I was not so familiar with web components but it isn't necessary ;).

malcom1120:
@FranckyC  - Thanks for the comments.  My other job has slowed  me from making the changes.  I hope to get them today. One question I had - How do you make a import dynamic?  I haven't done that before so if I had an example I could make that change.

FranckyC:
Np, a dynamic import is a Webpack functionnality. It look like this (example of a property pane control):
```
        const { PropertyFieldCollectionData, CustomCollectionFieldType } = await import(
            /* webpackChunkName: 'search-property-pane' */
            '@pnp/spfx-property-controls/lib/PropertyFieldCollectionData'
        );
```

This way your library is not bundled with the solution and loaded dynamically only when needed.

malcom1120:
@FranckyC  - Let me know if these meet the standards.  It might be Monday before I can get back to it, but hopefully this works.

PR #265, opened 2020-05-13
Issue-264: Race condition while registering / loading custom renderers

jhholm:
Sets the global scope variable as an array in cases where **getRegisteredRenderers** is called before **registerRenderer**. Fixes race condition as global scope object variables share mutations.


FranckyC:
Thanks @jhholm !

PR #271, opened 2020-05-19
Added Batch call to get ShortCut URL for .url filetypes

malcom1120:
Added Shortcut URL to Search Result
Created Aync function to get Shrotcut URLs in batch
Updated Search Results to add Shortcut URLs for results that are URL FileType
Updated getUrl handlebar helper to return ShortcutUrl for URL FileTypes

### DIFF
--- a/search-parts/src/models/ISearchResult.ts
+++ b/search-parts/src/models/ISearchResult.ts
@@ -24,6 +24,7 @@ export interface ISearchResult {
   [key: string]: string;
   IconSrc?: string;
   IconExt?: string;
+  ShortcutUrl?:string;
 }
 
 export interface IRefinementResult {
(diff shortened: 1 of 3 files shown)


wobba:
I like it :) I'll leave it to @FranckyC to verify and merge.

FranckyC:
@malcom1120 pretty good stuff. I wasn't aware the URL was stored directly in the .url file :). The only thing I see is if MS decides to change this format because it is not document anywhere I think. Also batching also comes with performances considerations when a lot of results are displayed but I think in this scenario it won't be an issue. Good job!

wobba:
I wouldn't expect anyone to touch that old shortcut format 😅

malcom1120:
Thanks @FranckyC  & @wobba . Until I had a user that was trying to create a "link" from one of their sites over to another site and wanted that to show-up in search, I didn't know the Link document type even existed. Then when I played with it and found out the OOTB search didn't handle them, I thought about making sure this project did handle them. 

I'm hoping document links are a limited used feature so the batching shouldn't become an issue, but thanks for pointing it out.

FranckyC:
@malcom1120 I approved too early :(. I get an error when the 'FileType' managed property is not present. Just make a check before using `toLowerCase` please.

image

jameelrrs:
Hi, We have been using PnP Modern Search WebPart V3 and planning to migrate to V4.  Now that we have been using .url files in one of the libraries and was working fine with V3 property "ShortcutUrl" in the Search Results template.  However, the same is not found on V4 version.  Are we missing anything here?  Is there some managed property or config to be enabled?  Please help

wobba:
@jameelrrs correct, feel free to log this as an improvement for v4 so we can track and maybe backport the solution.

Issue #274, opened 2020-05-20
Preview pane showing error text for some results in Simple List layout


AmitTyagi1:
**Version used**
3.12.1

**Describe the bug**
Preview pane showing below error text for results in Simple List layout where there are Folder result items.
 
`{"error":{"code":"itemNotFound","message":"The specified thumbnail set does not exist."}}`

![MicrosoftTeams-image](https://user-images.githubusercontent.com/33560478/82390516-c639c980-99f3-11ea-83b9-56527e3445b7.png)

**To Reproduce**
1. Add Search box, Search Results web-parts to the page
2. Configure the Search Results to show the results in Simple List layout
3. Enter some search term and press enter
4. In results look for the results for Folder names
5. Try to click on the preview image to open the preview pane.
6. See the error 

**Expected behavior**
It should show some default image or show text as - No preview available

**Desktop (please complete the following information):**
 - Chrome

AmitTyagi1:
Thanks @malcom1120 .

Unfortunately I don't yet have the development environment setup for SPFx projects. Can you please provide the .sppkg file with this fix? 

Thanks in advance.

malcom1120:
@AmitTyagi1 As a workaround until my bug fix gets merged into the package and released.

- Add IsDocument manage property to the returned properites
- Customize the default simple-list layout by changing  `{{#if (getPreviewSrc item)}}` with `{{#eq IsDocument 'true'}}`

AmitTyagi1:
Sure, I am going to try it right away. Thanks a lot.

AmitTyagi1:
@malcom1120 sorry it didn't worked for me. Do we have to make any changes in the `{{else}}` block too ?

malcom1120:
Attached is the full block for the preview container.  The preview container shouldn't even show if it is not a document.  I noticed yesterday that it was showing for Pages, Containers and ListItems.  My bug fix, fixes all of those.  This is the main part that corrects it.

`<div class="template_previewContainer ms-hiddenSm">
                                    {{#eq IsDocument 'true'}}
                                        {{#isnt FileType 'aspx'}}
                                        <div class="doc-container">
                                            <div class="img-container">
[...]

Issue #273, opened 2020-05-19
PDF preview shows explorer view instead


varunmshetty:
**Version used**
3.12

**Describe the bug**
When hovering over a pdf, the preview doesn't render the document view, but instead shows the explorer view of the folder where the document resides.

**To Reproduce**
This issue is not restricted to a single PDF. Tried this on over 20 different files.
Not restricted to a single tenant. Tried on a Developer and a Live tenant.
Search for a file and hover (click) on it to experience the behavior.

**Expected behavior**
The PDF preview must be shown and not the explorer view of its parent folder.

**Screenshots**
Screenshot attached.

**Desktop (please complete the following information):**
On Win 10, witnessed on Chrome (81.0.4044.138), Firefox (68.8.0 ESR), New Edge (81.0.416.72)

**Additional context**
The PDF preview of the out of the box search works just fine. Issue only with the PNP modern search.

![Search_PDF_Issue_DevTenant](https://user-images.githubusercontent.com/21065632/82385707-b8d70c00-9a00-11ea-9166-3741af908533.PNG)

malcom1120:
I had this same issue.  I'm looking at the code on this now.  It had an old custom template that still works with PDFs so I'm comparing that template and the new default templates.

wobba:
We'll look into it, and shouldn't be hard to fix :)

PR #275, opened 2020-05-20
Fixed Preview Bugs

malcom1120:
Added IsDocument to default managed properities
Added IsDocument check around the getPreviewSrc HB Helper so it only returns preview src for documents
Changed simple layout to match document card  preview for data-url in preview img (fixes PDF issue #273)
Changed handlebar block around preview to only show for documents that are not aspx filetype (Fixes container issue #274) - for some reason the if (getPreviewSrc item) always returns true even after fixing the helper to return "" for non preview-able items

### DIFF
--- a/search-parts/src/services/TemplateService/BaseTemplateService.tsx
+++ b/search-parts/src/services/TemplateService/BaseTemplateService.tsx
@@ -275,16 +275,20 @@ abstract class BaseTemplateService {
         Handlebars.registerHelper("getPreviewSrc", (item: ISearchResult) => {
 
             let previewSrc = "";
-            const nonSupportedGraphThumbnails = ["xls"]; //let's add more as we proceed
+            const nonSupportedGraphThumbnails = ["xls","aspx"]; //let's add more as we proceed
+            const nonSupportedPreviews = ["aspx"]; //let's add more as we proceed
 
             if (item) {
-                if (!isEmpty(item.SiteLogo)) previewSrc = item.SiteLogo;
-                else if ((!isEmpty(item.FileType) && nonSupportedGraphThumbnails.indexOf(item.FileType) === -1) && !isEmpty(item.NormSiteID) && !isEmpty(item.NormListID) && !isEmpty(item.NormUniqueID)) previewSrc = `${this._ctx.pageContext.site.absoluteUrl}/_api/v2.0/sites/${item.NormSiteID}/lists/${item.NormListID}/items/${item.NormUniqueID}/driveItem/thumbnails/0/large/content?preferNoRedirect=true`;
-                else if (!isEmpty(item.PreviewUrl)) previewSrc = item.PreviewUrl;
-                else if (!isEmpty(item.PictureThumbnailURL)) previewSrc = item.PictureThumbnailURL;
-                else if (!isEmpty(item.ServerRedirectedPreviewURL)) previewSrc = item.ServerRedirectedPreviewURL;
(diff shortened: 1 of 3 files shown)


search-parts/src/services/TemplateService/BaseTemplateService.tsx
```diff
@@ -275,16 +275,20 @@ abstract class BaseTemplateService {
...
-                else if (!isEmpty(item.PreviewUrl)) previewSrc = item.PreviewUrl;
-                else if (!isEmpty(item.PictureThumbnailURL)) previewSrc = item.PictureThumbnailURL;
-                else if (!isEmpty(item.ServerRedirectedPreviewURL)) previewSrc = item.ServerRedirectedPreviewURL;
-                else if (!isEmpty(item.ServerRedirectedURL)) previewSrc = UrlHelper.addOrReplaceQueryStringParam(item.ServerRedirectedURL, 'action', 'interactivepreview');
-                else if (!isEmpty(item.SiteId) && !isEmpty(item.WebId) && !isEmpty(item.UniqueID)) previewSrc = `${this._ctx.pageContext.site.absoluteUrl}/_layouts/15/getpreview.ashx?guidSite=${item.SiteId}&guidWeb=${item.WebId}&guidFile=${item.UniqueID.replace(/\{|\}/g, '')}&resolution=3`;
+                if (!isEmpty(item.IsDocument) && item.IsDocument.toLowerCase() == "true"){ //Preivew only for documents
+                    if (!isEmpty(item.SiteLogo)) previewSrc = item.SiteLogo;
```
wobba, line 283: Sites willnot have IsDocument I believe, so should be on top.

search-parts/src/templates/layouts/simple-list.html
```diff
@@ -59,16 +59,18 @@
...
-                                    {{#if (getPreviewSrc item)}}
+                                    {{#eq IsDocument 'true'}}
+                                        {{#isnt FileType 'aspx'}}
                                         <div class="doc-container">
                                             <div class="img-container">
-                                                <img class="document-preview-item img-preview" width="120" loading="lazy" src="{{getPreviewSrc item}}" data-src="{{getPreviewSrc item}}" data-url="{{#if ServerRedirectedEmbedURL}}{{ServerRedirectedEmbedURL}}{{else}}{{getPreviewSrc item}}{{/if}}" onerror="this.onerror=null;this.src='{{@root.utils.defaultImage}}';"/> 
+                                                <img class="document-preview-item img-preview" width="120" loading="lazy" src="{{getPreviewSrc item}}" data-src="{{getPreviewSrc item}}" data-url="{{#eq FileType 'pdf'}}{{#contains Path '-my.sharepoint'}}{{{ServerRedirectedEmbedURL}}}{{else}}{{{Path}}}{{/contains}}{{else}}{{{ServerRedirectedEmbedURL}}}{{/eq}}" onerror="this.onerror=null;this.src='{{@root.utils.defaultImage}}';"/>
```
wobba, line 66: Can you explain this change?

malcom1120: When looking at the preview for the tiles layout. It worked for PDFs.  I just copied the handlebar syntax that was in the preview URL for this one.  Since it worked, I figured it made since to make them the same.

search-parts/src/services/TemplateService/BaseTemplateService.tsx
malcom1120: All of the checks for returning a preview src are wrapped in the isdocument if so it is the first check. Is that were you were asking?

wobba:
Fixed with #280. I still want to keep showing front page image preview if present, and image preview for modern pages. Added a somewhat hacky error handler to catch errors on preview of images for pages where the preview handler fails.

PR #277, opened 2020-05-21
Persist selected values when filtering refiner options

joelfmrodrigues:
Small update on refiner options filter to always display selected options, even if they do not match the filtering text specified in the text box by the user


FranckyC:
Thanks @joelfmrodrigues, works well.

PR #288, opened 2020-05-27
Fix when searching people, as FileType is null & Sample using MS Graph Toolkit webcomponents

luismanez:
While working on another feature (gonna submit later), I found an issue in the _SearchService._mapShortCutUrls_ method. If you are searching for people, the following code raises an error, as FileType property is null for People result type:

```js
result.FileType.toLowerCase()
```

So I´m checking that property exists:

```js
if (result.FileType && result.FileType.toLowerCase() .......)
```


luismanez:
I tried to do a separate PR for the Graph Toolkit component sample, but fail miserably... There´s a second commit here with the sample. Apologies.

malcom1120:
Thanks @luismanez . I created this bug with a bug/enhancement fix that I did a few days ago. @FranckyC  caught it after merging and I was not able to get to it until today when I saw you had already corrected it.  @FranckyC - this corrects my to lower bug you found for PR #271 after merging.

luismanez:
Thanks @FranckyC. No worries, makes sense. Looking forward to see that new refactored version.
Cheers.

PR #292, opened 2020-05-29
Extensibility Upgrade

MakingThingsWorkToronto:
## Problem
- Support more than one extensibility library.
- Add Handlebars Helper extension.
- Lack of generalized / future proofed extension framework.

## Solution
- Added new project called search-extensibility containing common modules for use in extensibility scenarios, pulled models out of the search-parts project, refactored and added new.
- Added generalized extension framework patterned off existing extensibility & refactored code accordingly.

## Other changes
- Refactored extensibility service / library.
- Refactored Query Modifier, Web Component, Search Suggestion to leverage a common base instance class IExtension/IExtensionInstance.
- Added ExtensibilityEditor.tsx property pane control.
- Refactored searchBox/searchResults web part to use new ExtensibilityEditor.
- Refactored ISearchService to curry search parameters through a new interface / modified SearchService & MockSearch service for conformity.
- Modified BaseTemplateService to register handlebars helpers.
- Added extension context to include: search service, template service & web part context.
- Added new examples of all extension types / refactored search-extensibility-library.
- Updated documentation:
-- building the solution & linking search-extensibility project.
-- updated extensibility documentation.

## In Progress
- Translations (only english today).
- Improve package size (ExtensibilityEditor import for example).


MakingThingsWorkToronto:
Just a note, unfortunately I'm on the road for 2-3 days (moving). Can check comments but won't be able to make any changes until Sunday earliest. Please add any updates accordingly - this is primarily to be used in the refactored version but want to share code for testing purposes. Thanks EVERYONE!!

FranckyC:
Hi @MakingThingsWorkToronto , thank you for the PR. This is huge! I won't have time right know to review everything so sorry for that. Also, as you know, we are working on a refactoring currently and maybe this work should be integrated to the new base once released. I expect a first version mid summer. The both versions could also live side by side since there won't be a 1:1 match for features between the two versions at the beginning (that's why we don't accept new features). I'll let you know when the base will be ready.

MakingThingsWorkToronto:
Please don’t apologize! We all appreciate everything you’ve done! There are still improvements to make in this code so I will do that & hope to merge into the base when it is ready! Look forward to it! Matt

FranckyC:
After a quick review, here are my thoughts about the general extensibility feature. As you now, I've already done a similar job in my other solution and the plan was to open a part of this code for the PnP Modern Search v4 to stay aligned. However, it is interesting to see that we took a similar approach in the concept but with two different implementations. I tried to separate comments based on common topics. So, major differences between the two approaches:

## Architecture

Both use SPFx library components for common extensibility interfaces and custom extensions definition.

### Comments

In both cases, custom extensions and builtin ones follow the same interface and implementation pattern which is a good thing to avoid shortcuts and hacks for builtin ones. In normal time, I would prefer a public an npm package for extensbility interfaces like I did in my other solution but can be complicated to set it up in the PnP context for now.
	
## Extension definition

| Mine | Yours |
|------|-------|
| An interface and a base abstract class for a **each extension type** + a dedicated 'get' method (ex: `IDataSource` + `BaseDataSource` + `getCustomDataSources()`, `ISuggestionProvider` + `BaseSuggestionProvider`, etc.). Use of an SPFx service scope key property to register the default implementation. | An unique `IExtensionInstance` interface for all extension types + a base abstract class for each extension type and an unique 'get' method (ex: `IExtensionInstance` + `BaseSuggestionProvider` + `getExtensions()`). Use of a `extensionType` discriminant property to define the extension type.


### Comments

Not all extensions have an implementation class. As an example, for Handlebars, it is simpler to provide an isolated Handlebars namespace corresponding to the Web Part instance and let the developer do what he wants with it rather than providing interfaces for each Handlebars customization possiblities (helper, partials, etc.). The downside is you can't list them in the extensbility property pane panel.

## Extension instanciation

| Mine | Yours |
|------|-------|
| Usage of SPFx service scopes and dependency injection using service keys in the extension definitions. Web Parts calls the right method for the right extension when needed (ex: `getCustomDataSources()`, `getSuggestionsProviders()`, etc.). Extensions are loaded via dynamic imports reducing the bundle size. | Regular object creation via the `new()` operator using the extension class specified in the definition. Web Parts calls one single method `getExtensions()` to load all extensions for the library and a `filter()` method to get the right extension type according to the discriminant property.

### Comments

Service scopes allow more flexibility and control over object creation and lifecycle simplifying the code greatly by letting SPFx handle the class instance. It also povides an isolation between the main Web Part service scope and custom extensions. We can also share some services for consumers in that scope when needed. In both cases, we provide a `context` property to benefit from the Web Part context. In most cases, this will be used to get environement properties like current page URL, etc.. This shouln'd be used to get the Web Part service scope, for instance to get an instance of an SPHttpClient object or worse, a `ThemeProvider` instance`(which is different depending of the section where the WP belongs on the page).

## Extensibility registration

| Mine | Yours |
|------|-------|
| Loading by library manifest GUID + flag to enable/disable each library on-demand. No information displayed for exposed extensions for loaded libraries. Use of a PnP SPFx control for taht (DataCollection)| Loading by library manifest GUID. Information of available extensions in loaded libraries are displayed in the panel. Use of a custom control.

### Comments

The part where you list all available customizations is nice, maybe this part could be merged somewhere.

## Conclusion

First of all, this is a huge job and I really appreciate the effort you've put on. This is quite impressive! I feel embarassed because we took two different approaches that are, in fine, not really compatible. Also I don't want to throw away your code because you work is super valuable. When I designed this, I had more extension types in mind (data source, layouts, etc.) so that explain the differences I think. So what next? Here is what we can do:

- Use your version for the latest 3.x version of the PnP Modern Search.
- Use mine for the upcomming 4.x versions that will introduce major breaking changes anyway. In the 4.x the base is completely new and the architecture is very different and heavily based of service scopes. You could integrate gradually parts of your code in the new base when relevant.

Now, my concern is about change management and communication for users. They will have two ways of defining extensions between 3.x and 4.x. Not sure how we will approach that...The v4 is expected very soon (as a preview) so users won't have so much time to be familair with v3 new features. However, I think the two versions will co-exist for a short period of time (like few months), at least, during the time needed to get v4 at the same feature level of v3 (v4 is not feature ISO with v3 for now).

Hope this work for you. @MakingThingsWorkToronto  @wobba let me know what you think.

wobba:
I think it makes sense to merge into the v3 branch.

MakingThingsWorkToronto:
@FranckyC, @wobba ... As you know much of the work here was completed before the decision to refactor for version 4.

##Change Management
From a change management perspective, the additional dev complexity of the shared search-extensibility library may not be worth it for _the last minor release_. Agree adding to NPM & using install may solve that somewhat ... but it'll need to be managed by someone.

##Extensibility Definition
From a pure extensibility perspective, I see value in a simplistic import and the ability to scale out extension types via the getExtensions/filter methodology. It means fewer code changes when new extensions are introduced & a future proof pattern. We likely need to make some concessions to be _generic enough_ to handle future cases. My pattern abstracts the generic definition from the instance behavior which I liked. It sets us up to list in UI, enable/disable specific extensions etc. etc..

##Extensibility Registration
The ExtensibilityEditor property pane component: Validates the extensions can load successfully & provides feedback on specific cause of failure; Enforces standard naming to help troubleshooting; Displays what is available per web part to end users; Self documents extensibility implementation. 

##Extensibility Instantiation
I don't see a major difference here - I think we're both using Service Scope & given the library is loaded the performance consequences are minimal.

May be worth setting up a call to discuss if we really want to merge this. We could go back & forth forever in writing....

###Version 4
Given some companies have invested heavily in this, we need to be close to perfect on scripting the upgrade. As I mentioned I'd like to help - otherwise we're going to end up supporting multiple versions :)...

FranckyC:
As discussed, I'm closing it for now.

PR #318, opened 2020-06-18
CurrentCulture is used instead of CurrentUiCulture

engineererr:
**Problem**
This leads to having all labels in e.g. de-DE (set in SPO -> CurrentUiCulture)) but the search query is run in eg. en-US (set in Windows -> CurrentCulture).

This leads to no search results if I search for TaxFieldManagedProperty:"GermanLabel".

**Explanation**
Search query should be sent with CurrentUiCulture. In my case it would be 1031 (de-DE) but instead 1033 (en-US) is used.

image

**Info**
Documentation suggests using CurrentUiCulture as well: https://docs.microsoft.com/en-us/sharepoint/dev/spfx/web-parts/guidance/localize-web-parts#localize-the-web-part-manifest

And if I searched for other uses of CurrentCulture, I couldn't find another one in the project.

### DIFF
--- a/search-parts/src/webparts/searchResults/SearchResultsWebPart.ts
+++ b/search-parts/src/webparts/searchResults/SearchResultsWebPart.ts
@@ -231,7 +231,7 @@ export default class SearchResultsWebPart extends BaseClientSideWebPart<ISearchR
             }
         }
 
-        const currentLocaleId = LocalizationHelper.getLocaleId(this.context.pageContext.cultureInfo.currentCultureName);
+        const currentLocaleId = LocalizationHelper.getLocaleId(this.context.pageContext.cultureInfo.currentUICultureName);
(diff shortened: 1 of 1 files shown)


wobba:
Any reason the PR is work in progress? Can I flip it? :)

engineererr:
I created the pull request but was then not able anymore to prove that the fix is working...
I let you know tomorrow!

engineererr:
Search seems not to care about which culture we send with the post query.
If a sites default language is EN, and we tag a file with a multi-lingual term, search only picks up the EN label.

I wanted to give users a possibility so search labels in their language. Seems only possible if I translate the selected term to the GUID of the term and search with this one.

Still, I think the pull request can be merged to be more consistent.

wobba:
@engineererr Search does care about the culture, and is passed as en LCID. It's used for lemmatization.

Terms indexed does not care about language, as they are ALWAYS indexed using the primary label language. Hopefully this is something to improve in the future.

PR #319, opened 2020-06-18
Isolate result types rules for each webart instance

### LINKED ISSUE #316: Result types are conflicting if multiple search result webparts are on the same page
**Version used**
Ex: 3.12.1.0

**Describe the bug**
On a page where multiple search web parts are set up, configuring the result types on on the WP applies to all webparts.

**To Reproduce**

1. On a modern page, put two search result webparts
2. Configure both webparts to search for `*`
3. Configure both webparts to use the **simple list** layout.
4. On the first webpart, configure a result type rule:
   a. `Title` is not equals to 'aaaaa' (to ensure the rule applies to some document)
   b. Url of external template : whatever template you want
5. Publish the page

**Expected behavior**
I'd expect to have a custom template for only the 1st web part.

However, I see the external template applied to all webparts in the page.

**Screenshots**

 - Browser FF

**Additional context**
When switching to edit mode, I can see the correct render happens, seems like the render refreshed.

stevebeauge:
Fix #316

The idea is to register partials for each webpart instances.

The original `resultTypes` partial is replaced by a new one that rely on a helper that takes instance ID of the WebPart to select the actual partial to use.

The fix works in my case (a page with two search results webparts. One result type rule for one of the two). This should be reviewed I guess.

Hope that helped

### DIFF
--- a/search-parts/src/services/TemplateService/BaseTemplateService.tsx
+++ b/search-parts/src/services/TemplateService/BaseTemplateService.tsx
@@ -194,7 +194,7 @@ abstract class BaseTemplateService {
      */
     private registerTemplateServices() {
 
-        //https://support.microsoft.com/en-us/office/file-types-supported-for-previewing-files-in-onedrive-sharepoint-and-teams-e054cd0f-8ef2-4ccb-937e-26e37419c5e4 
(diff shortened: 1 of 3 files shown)


wobba:
@stevebeauge Can you update the PR as I added changes for the dynamic data fix to resolve the merge conflict? :)

stevebeauge:
@wobba this should be ok

PR #321, opened 2020-06-19
Adding german translation for webparts

ShaneBIS:
Translated the strings of the webparts to german.


wobba:
@ShaneBIS can you add translation for *ResultSourceIdDescription* as well which was recently added before I merge?

ShaneBIS:
@wobba Done.

PR #348, opened 2020-06-25
Updated Search Results and Search Query Box - with PageContext values/fields

kachihro:
Check the type of reference'd field - and split/use values accordingly


wobba:
You targeted the wrong branch, so please fix conflicts. You also missed som `decodeURIComponent`. You don't need to push package-lock.json.

Issue #332, opened 2020-06-23
Refinement does not filter results when on a result page greater than one


quarterhorse:
**Version used**
3.13.0 - May 2020

**Describe the bug**
When applying refinement based on a refinable string associated with a crawled property for taxonomy field only on page 1 of the results filtering is applied.  If on another page and a value is selected, no filtering is applied.

**To Reproduce**
1 - Issue a query resulting in multiple pages of results.
2 - Click on result page 3
3 - Select one value in the refinement panel to filter against a taxonomy field value
4 - The result set is not filtered
5 - Click on page 1 and the results are now filtered

**Expected behavior**
Clicking on a refinement value should apply filtering regardless of the current page

**Screenshots**

**Desktop (please complete the following information):**
 - Browser [e.g. chrome, safari] Chrome

**Additional context**

RamaKaryam:
We have the same problem. It used to work with separate pagination webpart.

wobba:
Agree. Applying a filter should reset to page 1 with that filter applied.

jindalv10:
@quarterhorse : I have fixed this issue. You can add following screenshot code in SearchResultsContainer.tsx file

![image](https://user-images.githubusercontent.com/13974138/85887829-78fd0480-b7b6-11ea-8232-43ef59db67b8.png)

wobba:
@jindalv10 you should do a PR 🙂

Jeroen00:
When can we expect the fix for this issue to be released?

wobba:
@Jeroen00 can you re-check with v3.14.2 as I think the fix was included there.

Jeroen00:
> @Jeroen00 can you re-check with v3.14.2 as I think the fix was included there.

I just tested 3.14.2 but unfortunately the issue still exists. I get search results with paging; I go to page 2 and filter/refine on a property but the results are not filtered and I remain on page 2. When I go back to page 1 the filter is also not applied, only works if I apply the filter when I already am on page 1.

wobba:
Fixed by #361

PR #361, opened 2020-06-30
[Bugfix] Revert to first page when search refiner value is selected

LeemansBE:
Hi,

There is a minor bug when you have searchresults with paging and search refiners enabled.
When you browse to the next page and apply a refinement filter the search results will become messed up.

This pull request will revert the search results back to the first page when a search refiner is applied.

Could you guys have a look? 😃 

Thx!
Bart


LeemansBE:
Fix for bug https://github.com/microsoft-search/pnp-modern-search/issues/332

xichro:
+1 Same issue.

wobba:
Thank you :)

wobba:
@LeemansBE I'm seeing an issue with refiners and verticals. If I pick a refiner on `Vertical 1`, then switch to `Vertical 2`. Then that looks good with refiners reset. If I switch back to `Vertical 1` then the first chosen refiner is applied, but not selected. Ideally a vertical switch should reset refiners and clear all state. I'm wondering if we need a connection between vertical and refiner web part, or if we can sort of work around it.

wobba:
Fixed with #424

Issue #302, opened 2020-06-05
Parameterized Search Page linked from Megamenu


melkinsco:
I realize this isn't per se an issue with the web parts, but wondering if anyone has solved this issue.  We are using a single search page, with search criteria passed to the page via the URL.  The problem we have is that, once the user clicks a link in the menu, and loads the page the first time, subsequent clicks of other links do not reload the same page with the new parameters.  Clicking the links does...nothing.  A refresh is required to click the new link.  Not a great experience for the user.

Thanks!

MakingThingsWorkToronto:
Can you elaborate on the specific link you’re referring to?

Pages in modern do not go through a full page refresh cycle but rather reload sections. 

If you want to force a full page reload on a custom hyperlink you can add an attribute called data-interception=‘off’.

There was an issue on a previous release where I saw behaviour similar to what you’re suggesting however it disappeared. To work around it I set the data-interception attribute to force a full page refresh. You likely want to make sure you’re on the most recent version (read release notes as there are breaking changes).

See more here: https://docs.microsoft.com/en-us/sharepoint/dev/spfx/hyperlinking

melkinsco:
Thanks for the quick reply!
Example of a link to Dental Benfits:
.../BenefitsSearch.aspx?BenefitTopic=Dental&data-interception="off"
Swap Dental for Vision and that would be a link to a page with Vision Benefits.  As you can see, the data-interception is set to off.
We're on version 3.12.0.0

3.12.0.0
--




3.12.0.0
--

wobba:
`data-interception=off` should be a parameter on the `<a>` tag, not part of href.

`<a href="some url" data-interception="off">text</a>`

sstrache:
Hello.  I work with melkinsco.   Thank you for the recommendation.   Unfortunately the modern menu structure only allows us to set an 'address' field (href value).  We can't add additional <a> tag attributes.  Have you seen the PNP web parts working in conjunction with the modern menu system using URL parameters?   Is there some modern site configuration we are perhaps missing that allows the menus to work?   

Thank you.

wobba:
@FranckyC could we implement something like https://docs.microsoft.com/en-us/sharepoint/dev/spfx/web-parts/guidance/intercepting-query-changes-in-webparts? Would we need special logic regarding if it's tied to a search box or not.

saregbesola:
[...]

Issue #230, opened 2020-05-04
Filters deep links - Many filters get selected issue


mahmoudhamed:
**Version used**
3.10

**Describe the bug**
I am using the Filters deep links feature to create a topic page experience. 
Currently I have many news pages tagged with terms from a term set called Topic. I created a managed metadata field in the site pages library to tag the pages and the field support multi-values. I am creating now a deep link to filter only for a single topic "Digitalization" for example but because also the pages are tagged with other topics these are also selected by default.

**Expected behavior**
Only the Digitalization filter should be selected.

**Screenshots**

![image](https://user-images.githubusercontent.com/7877441/80975646-a8306e80-8e22-11ea-8d92-d6af377ee31b.png)
![image](https://user-images.githubusercontent.com/7877441/80975702-b9797b00-8e22-11ea-94ff-2f98f0950b5e.png)
![image](https://user-images.githubusercontent.com/7877441/80976152-5cca9000-8e23-11ea-971c-8778b96e8c8d.png)

**Page field value example**
![image](https://user-images.githubusercontent.com/7877441/80976542-e0847c80-8e23-11ea-974d-f102fcd1d784.png)

**Desktop (please complete the following information):**
 - Browser: chrome

MakingThingsWorkToronto:
Appears the issue relates to line 1774 in SearchResultsWebPart > _mapDefaultSelectedFiltersToRefinementResults:

`
} else if (StringHelper.longestCommonSubstring(updatedSelectedFilterValue.RefinementToken, refinementValue.RefinementValue) && updatedSelectedFilterValue.RefinementToken.indexOf("range") === -1) {
// Means the provided condition in URL is an FQL expression so we try to guess the corresponding refinement results using the text value contained in the expression itself
`


For the above use case modifying the above if to the following resolves the problem.

`} else if (updatedSelectedFilterValue.RefinementValue.indexOf(refinementValue.RefinementValue) > -1 ) {
`

Having said that, I'm not sure I fully understand the intention behind the longestCommonSubstring comparison so unsure if this causes a regression.

FranckyC:
Hi @mahmoudhamed , maybe this would help to understand the behavior https://microsoft-search.github.io/pnp-modern-search/search-parts/search-results/#how-default-selected-values-are-set-in-the-search-filters-wp. You can try to use a KQL expression instead of the FQL value, in your case, this would be:

```
filters=[{"n":"RefinableString05","t":["Digitalization"]}]
```

[...]

PR #394, opened 2020-07-19
Fix URL change behavior #302

MakingThingsWorkToronto:
Resolves Issue #302 

There is a comment by @saregbesola about the configurable masthead search which I see as a bug not an enhancement. Given OOTB features have changed, I do see this as the bug that needs to be fixed ... also fixes the mega menu issue as well.

The implementation is clear cut and uses the guidance @wobba provided (with a little bit more cleanup): https://docs.microsoft.com/en-us/sharepoint/dev/spfx/web-parts/guidance/intercepting-query-changes-in-webparts

I know we're not accepting enhancements, #302 may be an enhancement but the masthead search box should work when we are on the configured search results page as well.

**_As a note_**, the search results web part is only handling this when the query keyword data source is the Page Environment. The search box web part is always handling the url change.


search-parts/src/webparts/searchResults/SearchResultsWebPart.ts
```diff
@@ -177,6 +182,8 @@ export default class SearchResultsWebPart extends BaseClientSideWebPart<ISearchR
             this._placeholder = Placeholder;
         }
 
+        console.log("rendering");
```
wobba, line 185: Please use spfx logging if needed,

MakingThingsWorkToronto:
Done! Also added a fix for #230

wobba:
Awesome!

PR #399, opened 2020-07-21
Added ability to set pnp-fabric-icon error image

MakingThingsWorkToronto:
Added the ability to set pnp-fabric-icon error image.

The error image can be used when the url passed to data-image-url fails to load. It does not work in scenarios where a data-icon-name is passed as the onLoadingStateChange callback is not fired.

As this is an enhancement please feel free to reject.


wobba:
Thank you (and you need to start targeting the dev branch ;) )

Issue #341, opened 2020-06-24
Refinement panel does not update when selecting a different vertical


quarterhorse:
**Version used**
3.13.0 - May 2020

**Describe the bug**
Using a modern search web part based page, when issuing a query with results and a populated refinement panel (to include taxonomy refinements), when the results are filtered the filtering is reflected in the results, when clicking on another vertical option (e.g. people), the refinement filters are not updated.

**To Reproduce**
1. Have indexed content that uses refinable managed properties
2. configure search results to use a refinement panel and verticals
3. Issue a query using one vertical that results in the refinement panel being populated
4. Filter results with the refinement panel options
5. Select a different vertical
6. Notice the refinement panel doesn't clear, reflect any refinables relevant to the new content, etc.

**Expected behavior**
When moving between different verticals, the refinement panel should update with the relevant refinable managed properties.  Refinables relevant to the new results should display, refinables not relevant to the new results should be deselected and removed, and, potentially, if current filtered values are applicable, the results are filtered by them.

**Screenshots**

**Desktop (please complete the following information):**
 - Browser [e.g. chrome, safari]  Chrome

**Additional context**

quarterhorse:
It may be working as expected - can someone quickly confirm?  I closed the browser and reopened it to my test page after over 30 minutes of checking and double checking by editing and saving as a draft.  My original test case used all three OOTB verticals and I added two more:

- Hub with a query template {searchTerms} DepartmentId:{{PageContext.hubSiteId}}
- Enterprise with a query template {searchTerms}

I was searching for content and refining it, but when I was selecting People, the refiners were showing the original selected values from both Hub and Enterprise even when the refiners were not relevant to the results.

Maybe what I saw was an artifact from editing and saving as draft?

wobba:
I think when we switch verticals we should clear any refiner state. It's not fully cleared right now.

Georg-Trixl:
The other bug is, when a refiner is selected, only the result numbers of the current vertical is reduced - the others do show the same result count, even when the other vertical could apply the same refiner. (We have an All vertical and Documents vertical, every Document shows also up in the All vertical - for that scenario one filter is suitable for both verticals)

wobba:
Which is correct - you cannot expect a refiner to apply to other verticals imo. I do see a bug still for resetting the refiner properly on  vertical switch which I'm looking into.

wobba:
Resetting refiners work now correct for all cases I believe - fixed with #424

[...]

PR #424, opened 2020-08-12
Fix/341

wobba:
Fix for #341


Jeroen00:
@wobba thanks! when do you expect this fix will be released in a new version?

wobba:
Most likely I'll package up a new version in the next week or so.

Issue #422, opened 2020-08-10
Person Refiner - Cannot read property 'toLowerCase' of undefined


thiagottss:
**Version used**
3.14.2.0

**Describe the bug**
Refiner/Filter webpart crashes when click to expand Person (using Person template) results.

**To Reproduce**
Click to expand Person. It started to occur after forcing the library to re-index. Before doing this action, it was working for newly created documents (freshly indexed).

**Expected behavior**
Expand and show Person template filters to choose.

**Screenshots**
Console Error Message:
react_devtools_backend.js:2273 TypeError: Cannot read property 'toLowerCase' of undefined
    at search-refiners-web-part_ec2cb2a12ad5aa9359f70b494aec5263.js:1
    at Array.filter (<anonymous>)
    at search-refiners-web-part_ec2cb2a12ad5aa9359f70b494aec5263.js:1
    at Array.map (<anonymous>)
    at t.render (search-refiners-web-part_ec2cb2a12ad5aa9359f70b494aec5263.js:1)
    at oa (sp-pages-assembly_en-ie_997ef04e722fc5f2ed093549634082ff.js:1)
    at aa (sp-pages-assembly_en-ie_997ef04e722fc5f2ed093549634082ff.js:1)
    at Oo (sp-pages-assembly_en-ie_997ef04e722fc5f2ed093549634082ff.js:1)
    at ko (sp-pages-assembly_en-ie_997ef04e722fc5f2ed093549634082ff.js:1)
    at bo (sp-pages-assembly_en-ie_997ef04e722fc5f2ed093549634082ff.js:1)

**Desktop (please complete the following information):**
 - New Edge browser.

**Additional context**
Using React Dev Tools I could export these data form the component:

`{
  "webPartTitle": "",
  "availableRefiners": [
    "{FilterName: \"Created\", Values: Array(6)}",
    {
      "FilterName": "RefinableString06",
      "Values": [
        {
          "RefinementCount": 57,
          "RefinementName": "RefinableString06",
          "RefinementToken": "\"ǂǂ53617261682e486f67616e404344424176696174696f6e2e6165726f207c20536172616820486f67616e207c2036393341333032333245363637433644363536443632363537323733363836393730374337333638364636373430363336343632363136433636324536333646364420693a30232e667c6d656d626572736869707c73686f6740636462616c662e636f6d\"",
[...]

PR #430, opened 2020-08-17
fix to handle userInfo is undefined when user is disabled

thiagottss:
https://github.com/microsoft-search/pnp-modern-search/issues/422


wobba:
Thank you :)

PR #507, opened 2020-10-04
Added handlebars helper filterArrayByProperty

anoopt:
## Problem

_What problem are you trying to solve?_

We can filter the search results on managed property values using KQL however sometimes we might not get the exact filter we need using KQL. Example _filter the results which have first character of the Title between A-H_ . 

The existing Handlebars helpers do have a `regex` helper however that cannot be used directly in filtering an array.

## Solution

_How did you solve the problem?_

A new helper named `filterArrayByProperty` has been added in this PR. The helper function accepts an array, name of the property and a regular expression. Based on the property and the regular expression, the function returns a filtered array that matches the regex specified for the property.

## Example use case

Say we want to group the results in 3 groups of alphabets - A-H, I-P and Q-Z. In that scenario we can use this helper to get the filtered items for these groups i.e. get the items which have the first character of Title between A-H, the I-P and then Q-Z.

The results would be as shown in the image below (SharePoint List on the left and the PnP Search Results webpart on the right)

image

Template needed for the example above

```html
             <ul>
                <b>A-H</b>
                {{#each (filterArrayByProperty items "Title" "^[a-hA-H]")}}
                    <li>
                            <span>{{Title}}</span>
                    </li>
                {{/each}}
                <br />
                <b>I-P</b>
                {{#each (filterArrayByProperty items "Title" "^[i-pI-P]")}}
                    <li>
                            <span>{{Title}}</span>
                    </li>
                {{/each}}
                <br />
                <b>Q-Z</b>
                {{#each (filterArrayByProperty items "Title" "^[q-zQ-Z]")}}
                    <li>
                            <span>{{Title}}</span>
                    </li>
                {{/each}}
            </ul>
```

## Other notes

The file `IFilterLayoutProps.ts` has a property named `eLanguage` which is causing an issue. Hence during development of this PR that property had to be changed to `language`.

The scenario explained in example above can be solved using the combination of `regex` and `startsWith` helpers. However, that would be slightly complex approach. 

Apologies if I have overthought this and missed a different way of solving the problem.

### DIFF
--- a/docs/search-parts/templating.md
+++ b/docs/search-parts/templating.md
@@ -216,7 +216,7 @@ Setting | Description
 `{{#group items by="owstaxidmetadataalltagsinfo"}}` | Group items by a specific results property. See [https://github.com/shannonmoeller/handlebars-group-by](https://github.com/shannonmoeller/handlebars-group-by) for usage.
 `{{#getAttachments LinkOfficeChild}} <a href="{{url}}">{{fileName}}</href> {{/getAttachments}}`| Get Attachments is Handlebars block helper intended to be used with the context variables it provides like this (url and filename) in the example.The helper is intended to be used only with the LinkOfficeChild managed property wich is the default managed property for list attachments.
 `{{getPageContext "property"}}` | Retrieve SPFx page context data to show in the template. See [https://docs.microsoft.com/en-us/javascript/api/sp-page-context/pagecontext](https://docs.microsoft.com/en-us/javascript/api/sp-page-context/pagecontext) for possible values. Example `{{getPageContext "user.displayName"}}` `{{getPageContext "cultureInfo.currentUICultureName"}}`.<br>**Note: Casing matches the object model.**
- 
+`{{filterArrayByProperty items "property" "regex"}}` | Filter items based on a property's value using regular expression. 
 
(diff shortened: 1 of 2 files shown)


wobba:
Curious if you cannot solve the same using `#each` and `isMatch`.

```
{{#each items as |item|}}
	<li>
		{{#if (isMatch Title "[i-pI-P]*")}}
		{{Title}}
		{{/if}}
	</li>
{{/each}}
```

anoopt:
Thank you for your response Mikael. Yes that will give me the required results. I should have asked this as a question instead of submitting a PR!

Do you mind answering this question please? - If I have to get the results categorized by group of alphabets (A-H, I-P and Q-Z) with the headings at the top of each group (as shown in the screenshot), then I have to iterate over the items 3 times is that right or is there way of getting the results with headings by iterating over the items only once?

wobba:
@anoopt You would need to loop once per section, yes. You might can sort them first and then loop, but I think the first approach is easier. For v4 we are thinking to allow you to deploy custom helpers yourself as a library. Which makes it easier to extend when needed.

anoopt:
Thank you for your response Mikael. That is very helpful. Please ignore this PR. 

Like I said earlier, "Apologies if I have overthought this and missed a different way of solving the problem.".

wobba:
No problem 🙂 We're happy for improvements so feel free to suggest something in the future. To avoid doing work which might not be needed you can always ask and suggest before a PR as well.

Issue #31, opened 2019-12-06
The "DynamicData" object cannot be used because it has been disposed


PooLP:
## Category

- [x] Bug

Hi,
I have a weird bug on first release tenant. 
I haven't tested on GA tenant.
I tested only with Chrome.
It's last release of pnp-modern-search (3.8)

## Expected or Desired Behavior

If you have many page who use pnp-modern-search webparts and if you navigate between her with partial load you have this error in console :

`The "DynamicData" object cannot be used because it has been disposed`

Yesterday, I also had this error that makes crash this webparts, but I do not reproduce it this morning :

`Source has already been initialized`

## Steps to Reproduce

- open chrome console
- create two page and configure pnp-modern-search with all search webpart (searchbox, refiners ...)
- create two nav access in header nav bar
- navigate several times alternately between these two pages
- see the console

Thanks

FranckyC:
Hi @PooLP, thanks for submitting this issue. When this issue occurs, does it break the connections between Web Parts on the page?

FranckyC:
Ok I'm able to reproduce it. I'll see what could be the cause. Thanks.

FranckyC:
@PooLP Should resolved by #34 . Can you confirm?

PooLP:
I'm confirm, it's resolved!

cpluceDev:
this happens when navigating from modern search pnp web parts and OOTB Microsoft(Modern) search from suite bar  https://github.com/microsoft-search/pnp-modern-search/issues/537

waynel2:
Hi there,

Just wondering if there has been any further troubleshooting or if anyone can assist to resolve the above issue?

A colleague of mine has just raised the same issue under under thread (https://github.com/microsoft-search/pnp-modern-search/issues/560 ) hence if anyone has any ideas on how to resolve this it would be great to hear from you!

Thanks

Wayne

wobba:
@cpluceDev @waynel2 should be fixed in the dev branch now.

waynel2:
@wobba - that's for the heads up.  I'm only a BA though will let our developers know so hopefully they can deploy the latest version.

PR #542, opened 2020-11-09
[PnP Modern Search v4] New code base

FranckyC:
@wobba @MakingThingsWorkToronto Here is the new code for the PnP Modern Search version 4. Please take a look at the solution and at least deploy it to your tenant/site collection.

If everything OK, I will merge to develop then main branch + create a new release.

The IDs (solution + WPs) are not the same with v3 so you can use them side-by-side.

Enjoy!


waynel2:
Hi there,

Please can you advise if the below issue will be addressed in v4?

https://github.com/microsoft-search/pnp-modern-search/issues/560
https://github.com/microsoft-search/pnp-modern-search/issues/31

Thank you

Wayne

search-parts/src/webparts/searchFilters/SearchFiltersWebPart.ts
```diff
@@ -92,6 +93,12 @@ export default class SearchFiltersWebPart extends BaseWebPart<ISearchFiltersWebP
    * The available web component definitions (not registered yet)
    */
   private availableWebComponentDefinitions: IComponentDefinition<any>[] = AvailableComponents.BuiltinComponents;
+  
+  /**
+   * Custom configurations
+   */
+  private availableConfigurations : IConfigurationDefinition<any>[] = [];
+  private activeConfigurations : IConfiguration[] = [];
```
MakingThingsWorkToronto, line 101: Extensibility is not exposed in this web part. Was going to chat with you about adding it. Would be nice to use custom configurations anywhere there is a template. May want to revert changes to this file if we don't add extensibility.

PR #628, opened 2020-12-17
Fix mapping attempt of missing aggregations on Graph objects

mbice:
When testing out Microsoft Search with a ServiceNow connector, the query succeeded in pulling data, but failed when evaluating the response payload. Using my connector, the Graph query's result payload has no aggregations specified so that doesn't exist on the object at all. The code attempts to map aggregations in MicrosoftSearchDataSource.ts, which fails.

image
image
image

Fixed it by adding a check and only mapping aggregations if they exist on the response Graph query's payload.

Thanks!


wobba:
Thank you :)

PR #658, opened 2021-01-12
Add queryTemplate option to searchbox

YannickRe:
Add the ability to transform the searchbox entry before sending it off to another page. Useful when you want to always enrich the search query the user will enter.


wobba:
We'll get this processed - just a bit busy :)

YannickRe:
Don't worry, it's not urgent. I just needed this for a customer requirement and decided to submit, up to you to see if it'd fit the main product 😃

FranckyC (commented):
I'm not really comfortable with this change honestly as it introduces an implicit relationship between the search box and the SharePoint search engine. The search box should remain agnostic and should only get user keywords as free text. The query template is only relevant in a SharePoint search context. What if we introduce an other data source in the future like Azure Search? In this case, the query template setting becomes useless and creates confusion.

I'm sure your customer has good reason to do this but for a general purpose and future evolution I prefer leave the search box untied to any data source. Hope you understand.

FranckyC:
Sorry for the delay @YannickRe. quite busy here ;).  Check my review.

YannickRe:
@FranckyC I understand, but I'm not sure I agree 🙂.

It is named "queryTemplate" but it applies the token transformation before sending the result to the search results web part OR a new page. This means you can have a queryTemplate "{searchTerms} RefinableString00:{Page.MyMetadataField}", and when you enter a search term in the box it will be added in place of "{searchTerms}" and all other tokens will be parsed too. Afterwards, the total resulting string will be sent to the search results web part as a query.

So unless the tokenservice is specific to SharePoint (which I don't think it is?), I don't see the implicit relationship. I do agree that the naming is confusing, but I just wanted to align with what exists already.

wobba:
@YannickRe which scenarios would it not work to have the tokens in the result web part, but you want it in the search box?

wobba:
@YannickRe I guess you're looking for `Page` properties in particular to pass over right? To add additional context on the redirect?

YannickRe:
Correct, I need contextual information from the source page to be passed to the destination search page

FranckyC:
Ok so in this case, why not tokenize the input query text directly and avoid the 'queryTemplate' parameter which is, I persist on that point a SharePoint search concept?

wobba:
I think it makes more sense to have an option builder where you state the query param, and pick a Page token value for the redirect scenario. A bit more work, but more flexible and intuitive. This allows the data to be picked up by any dynamic data enabled part as well. Agree, disagree, up for it? :)

image

FranckyC:
Or apply tokenization in the PageURL field instead so you could put any context value here as query string params? For all other scenarios, this can be done in the search results wp

wobba:
Adding tokens in the URL field is a good idea. @YannickRe if you make that change, and add documentation to it as well to https://microsoft-search.github.io/pnp-modern-search/usage/search-box/ I think we have a winner. And specify which tokens from https://microsoft-search.github.io/pnp-modern-search/usage/search-results/tokens/#page-tokens are applicable.

FranckyC:
I'm OK to apply tokenization only in the Page URL field and remove the query template parameter. Then a good documentation about what tokens are supported in this case and we are good to go I think.

YannickRe:
I don't think it will solve my use case. Passing addition querystring parameters which are parsed, will mean I need to read them separately from the query string on the results page. I want to offer the opportunity to remove them for a subsequent search on the result page.

My need:
- I need this to be going from source page to search page: `/SitePages/search.aspx?q=myTextQuery RefinableString00:MyValueFromPageMetadata`
- Search Page will read `q` as the complete search query, and show it in the search box on the search.aspx page
- Users now can adjust the query on the result page, keeping the additional RefinableString00 query part or removing it and doing a completely new query on the results page.

Everything that is entered in the search box is in the current implementation the value that will go into the `q` parameter string. My change to use query template with `{searchTerm}' as default is a non-breaking change. This will work for all existing installations of the web parts. It will parse the queryTemplate, inject the search term from the box in place of `{searchTerm}` and use that is value for `q`. It is VERY flexible, as you can also do things like `fixedValue RefinableString10:{searchTerm}` as your queryTemplate.

Your suggestion would also have no impact on existing installs, but would be using this URL structure: `/SitePages/search.aspx?q=searchValue&param1={Page.FieldValue}`. The search results web part need to read those queryString parameters in the queryTemplate field. It removes all flexibility of end users to remove that part of the query for a subsequent query.

What would work for me, but is a breaking change: change the URL field from `/SitePages/search.aspx` to  `/SitePages/search.aspx?q={searchTerm} RefinableString00:{Page.FieldValue}` and tokenparse it like this. It will break existing installations AND the option to choose between query hash and querystring for `q`.

I stand with my opinion that my implementation is more flexible and less breaking, but only with a potential naming clash that you associate with SharePoint. I want to achieve a query transform on the entered search term BEFORE sending it to another page as `q`. That's not just useful for SharePoint.

FranckyC:
Thanks for the requirements clarification @YannickRe. In the solution I propose regarding the URL field, tokens would be resolved **before** sent to the target page. So if you have a page **A** with a search box redirecting to a page **B** with the following format `/SitePages/pageB.aspx?q="{inputQueryText} RefinableString00:{Page.Field1}"`, the page **B** will get the resolved values from `searchBoxQuery` and `Page.Field1` in the context of page **A** through the _q_ query string parameter. If you configure the search box on page **B** to use dynamic data source with the _q_ parameter, users will be able to change the query afterwards. There is no need for the search results to parse this parameter in this scenario.

I'm not comfortable to introduce an option that is only targeted for a specific use case: send a transformed query to an other page. We can do this in the URL field in my opinion like this field is your query template in the end.

To ensure existing implementations will still work, if no explicit `{inputQueryText}` token is defined in the URL (corresponding to "manual" mode), we send the whole input text in the _q_ parameter (the "default" mode). Also you have to be careful with the token service because instances are not shared between search box and search results. It means you have to define and set first the token values you want to support in the search box Web Part (like `inputQueryText` in this case).

Hope this make sense :-)

YannickRe:
I understand your suggestion but I believe this will still be problematic in many use cases.

1. You say you aren't comfortable to introduce an option that is only targeted for sending a transformed query to another page: but isn't this exactly what your solution does? You are adding the token transformation to the URL, thus in case where you send the query from the search box to a search result on the same page the transformation isn't applied.
My option (and current implementation) transforms the query before it "leaves" the search box works for search results web parts on the same page as well as on a different page.

2. Your suggestion would require people to add `?q=`or `#q=` to their search.aspx page URL. It will break the following functionality where this is configurable:
PnPModernSearch - Copy

3. You say: "we can do the transformation in the URL as this is your queryTemplate in the end". I agree that it is conceptually the same, but why do you insist on making it LESS user friendly by putting it in the URL text box while adding a new box:
- is more user friendly
- has less code impact
- has less risk on regressions

4. This is how mine looks, please consider it again. With redirect:
PnPModernSearch
Without redirect:
Capture

FranckyC:
Ok so:

- _"My option (and current implementation) transforms the query before it "leaves" the search box works for search results web parts on the same page as well as on a different page."_.

On a same page scenario, this is not the purpose of the search box to resolve tokens, but the search results Web Part. I really don't see the use case here, sorry. I think it is misleading.

- "Your suggestion would require people to add ?q=or #q= to their search.aspx page URL. It will break the following functionality where this is configurable:"_

That would require some parsing to see if a query string has been manually set. Not the best approach I agree on that. In this case, why not adding the "Query template" field only when the **Send query to a new page** is checked. The value set here would be append to the query string or hash according to the existing setting. The default value for this field should be `{searchTerms}` corresponding the current search box value (even if I'm not a fan of the token name, that's fine). Also, to avoid confusion, maybe rename this field to something more accurate like "Query string template".

Last thing: you should use the `setTokenValue` method to set the `{searchTerms}` value every time the search box input is updated instead of a regex + string `replace`. See https://github.com/microsoft-search/pnp-modern-search/blob/3f25fe253e6829fe46c326243706c0050c869937/search-parts/src/webparts/searchResults/SearchResultsWebPart.ts#L1737. Then use the `resolveTokens` method when the query is sent through dynamic data.

Would that work?

YannickRe:
Ok, I think this can work out. Made the first changes already, and I will get back with an update to the PR.
I will change `searchQueryTemplate` to `inputTemplate`, as it doesn't have the SharePoint connotation.

FranckyC:
Don't forget the documentation @YannickRe  ;)

YannickRe:
For sure not, I’ll get the code first so we are aligned on that first.

YannickRe:
@FranckyC Mind reviewing the code? Resulting property pane is now:
image

FranckyC:
Hi @YannickRe I'll do this week for sure. Sorry for the delay, quite busy with a customer project.

FranckyC:
@YannickRe Tested and works well. Can you update the documentation about this feature usage? Then we're good to go.

YannickRe:
@FranckyC docs are done, hopefully to your satisfaction :D

FranckyC:
All good thanks @YannickRe !

PR #758, opened 2021-02-09
Added information about approving access requests

sympmarc:
I found it confusing to follow the installation steps, so I added a section on approving the access requests.


FranckyC:
Thanks @sympmarc ! Your changes should be visible soon in the live documentation.

PR #785, opened 2021-02-19
Added danish translation

allanhvam:
Added Danish language support to the webparts.


FranckyC:
Thanks @allanhvam, can't really validate if all terms are correct so I will trust you on this ;)

PR #793, opened 2021-02-22
Feature/v4mtw

MakingThingsWorkToronto:
Added className properties to filter components for styling.


MakingThingsWorkToronto:
getPageContext handlebars helper re-added to TemplateService.ts to align with v3.

MakingThingsWorkToronto:
getAttachments handlebars helper re-added to TemplateService.ts to align with v3. I believe all the custom helpers that were in v3 are now in v4. @FranckyC, these were added by @wobba I believe while you were on hiatus.

PR #936, opened 2021-04-08
Fixed problems with localization resource files

s-KaiNet:
I found some inconsistencies in localization resource files - some labels were removed, some migrated to new nodes, one label was missing. This PR fixes all that issues.   

To summarize the changes: 
- _commonStrings.d.ts_:
	- `Sort.SortableFieldManagedPropertyField` and `Sort.SortableFieldDisplayValueField` are not used anymore in solution (they are missing in corresponding `{locale}.js` files and no usages were found). So I removed them altogether from `commonStrings.d.ts`
	- `DetailsList.SupportHTMLColumnLabel` is not found in `commonStrings.d.ts`, so I removed corresponding properties from `{locale}.js` files
- _searchResults/loc/mystrings.d.ts_:
	- `LayoutPage.ManageDetailsListColumnLabel` was moved to the `common.d.ts` localization files, thus I removed this definition
	- `ResultTypes.DialogButtonLabel` is not defined in `searchResults/loc/mystrings.d.ts` (actually it was moved to `LayoutPage.DialogButtonLabel`), thus was removed from corresponding {locale}.js files
- _searchBox/loc/mystrings.d.ts_:
	- `da-dk.js` is missing `SearchBoxSettingsGroup.QueryInputTransformationLabel` property. I'm not sure what is the right Danish translation, so I just added an English version to fix the problem. If you could tell me the right translation in Danish for "Query input transformation template", I will update the PR.   

How I found all those issues? Recently I've published a VSCode extension called [SPFx Check Locale](https://marketplace.visualstudio.com/items?itemName=s-kainet.spfx-check-locale), which dynamically checks SPFx solutions for potential problems with localization files. It's also available as a Nodejs module, so you can integrate it to your build pipeline and fail a build if there are issues with resource files - [check it out an example here](https://github.com/s-KaiNet/spfx-check-locale).

### DIFF
--- a/search-parts/src/loc/commonStrings.d.ts
+++ b/search-parts/src/loc/commonStrings.d.ts
@@ -145,8 +145,6 @@ declare interface ICommonStrings {
           SortPanelSortDirectionLabel:string;
           SortDirectionColumnLabel: string;
           SortFieldColumnLabel: string;
-          SortableFieldManagedPropertyField: string;
-          SortableFieldDisplayValueField: string;
           EditSortLabel: string;
           SortInvalidSortableFieldMessage: string;
           SortFieldColumnPlaceholder: string;
@@ -323,9 +321,8 @@ declare interface ICommonStrings {
       }
     }
 }
-  
+
 declare module 'CommonStrings' {
   const strings: ICommonStrings;
   export = strings;
 }
-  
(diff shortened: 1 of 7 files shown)


s-KaiNet:
Maybe it makes sense to exclude `loc/*.js` from duplication checks since SonarCloud reports a lot of duplications there, however, this kind of duplication doesn't add any problems to the code. And probably for `.html` templates as well.

wobba:
Looks good :)

PR #974, opened 2021-04-30
Fixed small issues in Swedish resource files

s-KaiNet:
I noticed some labels in Swedish localization were obsolete and removed them in order to maintain consistency across resource files.


wobba:
Thank you :)

Issue #720, opened 2021-01-28
Filter/Refiner panel not displaying filtered fields


smitman3405:
I've added terms to the term store and created managed metadata columns to those items in the SitePages library.  I've tagged some pages with the terms via page properties but I'm not getting any results in the Filters webpart.  I've even gone as far as mapping the fields in the Managed Properties in Search in SPO Admin (RefinableString00).
I've also tried adding a choice column to the list but that choice field does not show up in the choices when setting up the custom filters in the Filters Webpart.
I'm using the latest PNP Modern Search release.  
What am I missing???

![image](https://user-images.githubusercontent.com/10451710/106192080-b75a0800-6179-11eb-85c3-2dc8e4c7374b.png)
![image](https://user-images.githubusercontent.com/10451710/106192272-f25c3b80-6179-11eb-85f9-d94ebd8f1abd.png)
![image](https://user-images.githubusercontent.com/10451710/106192361-13bd2780-617a-11eb-9fc7-7e3edaf872f1.png)
![image](https://user-images.githubusercontent.com/10451710/106192515-49621080-617a-11eb-9cbf-a25c7748ad9b.png)

FranckyC:
Hi @smitman3405, in the filter configuration, you must use `RefinableString00` as the filter field name. The screenshot you provided indicates this is not the case.  It the value doesn't appear in the dropdown, type it manually in the field and press "Enter".

smitman3405:
Thank you!!!  I never thought to type in the value, I thought I'd only have access to the ones in the drop down.

Now I have another question.  Do you know how to modify the search results template so just the values of the taxonomy items show and not the L0|#<guid>|
![image](https://user-images.githubusercontent.com/10451710/106266939-0003d680-61f7-11eb-9f3b-4c096ded00ee.png)

I'm guessing the Value#<guid> in the refiners is because I was including the actual ows_taxid_System_x0020_Tags in the crawl.  I've removed that and need to wait for the data to be crawled again before it will disappear.
![image](https://user-images.githubusercontent.com/10451710/106267021-16119700-61f7-11eb-9cf3-22f892196fc3.png)

wobba:
Closing as original question is answered. Label vs guid is covered in docs and other issues.

sympmarc:
@wobba I just ran into the label vs. GUID issue for taxomony fields, and I can't find anything in the docs to help me fix it. I know I've solved it before, but I'm coming up empty this time. Could you provide a link?

[...]

PR #984, opened 2021-05-04
Scenario-based docs: first pass

sympmarc:
As discussed in issue #720, I've created a start at some scenario-based docs. I intentionally didn't go very far so I could get your reactions before I put too much time into it.

My thought is the scenarios can be additive, listed in rough order of complexity. I fleshed out Scenario 1; the other two scenarios are just outlined:

* Scenario 1: Create a simple search page
* Scenario 2: Set up Managed Properties
* Scenario 3: Build a page showing content with filters which are Managed Metadata columns

The new docs are in the `scenarios` folder and the assets are in a subfolder, also called `scenarios`.

Let me know what you think!


wobba:
Great start! I'll merge this one, and we can iterate.

sympmarc:
Great! Let me know if you have any thoughts on the content. I can flesh out the other two scenarios. What others would be useful based on the questions you get?

sympmarc:
I got a handful of suggestions for scenarios from a tweet I did. I've added them in my local copy.

PR #1003, opened 2021-05-17
Added support for the next version of fast-serve.

s-KaiNet:
Fast-serve received an additional major update, which simplified a lot of things. This PR adds support for spfx-fast-serve@3.x version.    

## How to upgrade to SPFx 1.12 later on

When you're ready for the upgrade, then you just need to change the version of `spfx-fast-serve-helpers` from `~1.11.0` to `~1.12.0` (don't forget `~`, it's important).


wobba:
I trust you for the removed packages :)

s-KaiNet:
@wobba yeah.. it should be fine because it's the same set of deps I added way back before [here](https://github.com/microsoft-search/pnp-modern-search/pull/144/files) when I initially introduced the very first version of fast-serve. Now it's simpler and adds a lot smaller footprint to the solution. 

Additionally, I removed [`node-sp-auth`](https://github.com/s-KaiNet/node-sp-auth) since you don't use it anymore. To my memory, only the first version used this package.  

Also I verified that at least it builds in dev and prod modes, basic functionality for webparts works :wink:

PR #1009, opened 2021-05-18
[v3] Fix issue with persona template not loading

### LINKED ISSUE #1008: Persona refiner not displaying values when large list of values
**Version used**
Ex: 3.19.2

**Describe the bug**
Persona refiner not displaying refiner values when the list of values is very large.

I have debugged this and the issue happens within the `getUserInfos` method of the UserService when the component tries to batch query all users to get their account name and display name. The batch query never gets resolved when the number of batch request is very large, but works when the number is reduced.

As the getUserInfos service method only retrieves the account and display name, I believe this is not required as we have that information in the refinement value string already. There are already different places within the same control where the display name is being extracted from the refinement value string.

I will submit a PR with this approach as I have tested it and seems to be working as expected.

**To Reproduce**
1. Setup a refiner based on a people field
2. Ensure the query returns a very large number of refiner values for that people refiner (500+)
3. Persona refiner stuck on loading 

**Expected behavior**
Display refiner values

**Screenshots**
image

 - Browser [Edge (modern)]

**Additional context**

joelfmrodrigues:
Fixes #1008 

Please note that this change seems to completely remove dependency on UserService, but I have not removed it from the solution as it is currently passed as property between many components. I can update the PR to do this if required.

### DIFF
--- a/search-parts/src/webparts/searchRefiners/components/Templates/Persona/PersonaTemplate.tsx
+++ b/search-parts/src/webparts/searchRefiners/components/Templates/Persona/PersonaTemplate.tsx
@@ -148,26 +148,24 @@ export default class PersonaTemplate extends React.Component<IPersonaTemplatePro
 
   private async getUserInfos(props: IPersonaTemplateProps) {
 
-    const accountNames = [];
-
+    const userInfos: IUserInfo[] = [];
     props.refinementResult.Values.map((refinementValue: IRefinementValue) => {
 
       // Identify the login adress using Regex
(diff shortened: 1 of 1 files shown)


wobba:
You have changed to assume you always have display name + account name in the value. That's not necessarily the case. the code should cover if only an account name is present as well - and for those load the display name if needed in a batch.

joelfmrodrigues:
@wobba thanks, that definitely makes sense and I missed that case. Extracting the display name from the refiners is probably not an issue in the other parts of the code.
I will close this PR and approach the problem in the way you are suggesting.

wobba:
@joelfmrodrigues you can update your branch while the PR is pending as well :)

PR #1012, opened 2021-05-19
Only load persona card details if required

### LINKED ISSUE #1008: Persona refiner not displaying values when large list of values

joelfmrodrigues:
Fix #1008 

Only load details of users that don't have display name, instead of loading details for all users. Update array of users with retrieved display name

### DIFF
--- a/search-parts/src/services/UserService/IUserService.ts
+++ b/search-parts/src/services/UserService/IUserService.ts
@@ -1,4 +1,5 @@
+import { IUserInfo } from "../../models/IUser";
 
 export default interface IUserService {
-  getUserInfos(accountNames: string[]): Promise<any>;
+  getUserInfos(accountNames: string[]): Promise<IUserInfo[]>;
 }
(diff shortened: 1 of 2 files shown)


FranckyC:
Thanks @joelfmrodrigues Looks good ;)

PR #1014, opened 2021-05-21
Clear sorting

MarkusLanger:
After a re-sort has been done in a result web part, it can now be undone.

1.) Configure additional sorting in the result web part
2.) Select the term in the drop down
3.) Now a cross appears next to the drop down box to remove the custom sorting

image


wobba:
Thank you :)

PR #1016, opened 2021-05-22
Fix [V3] Apply button always active for Multi-value refinement item t…

### LINKED ISSUE #1015: [V3] Apply button always active for Multi-value refinement item template
**Version used**
Ex: 3.19.2

Apply button always active for Multi-value refinement item template, even if the current selection is the same as the refiners already applied. 

**To Reproduce**
1. Configure refiner using Multi-value refinement item template 
2. Select some values
3. Click Apply
4. After the selection is applied, confirm that apply button is still active and can be clicked again

**Expected behavior**
Apply button to be disabled when no refiner items are selected or when the refiners applied are equal to the selection (no changes made to selection after applying)

 - Chrome
 - Edge

joelfmrodrigues:
Fix #1015

Add a condition to disable Apply button for multi-value checkboxes when the refiners currently applied are equal to the ones currently selected in checkboxes.


wobba:
Thank you 🙂

PR #1057, opened 2021-06-09
Fix #1056

### LINKED ISSUE #1056: Incorrect encoding when connecting search box and search results
**Version used**
Ex: 4.2.2

**Describe the bug**
I often use the combo search box + search results WP, connected using dynamic values.

However, when there's a space in the query (maybe other special chars too ?), the query encode the space with a `%20`.

For example, typing `some term` in the search box results in the query `some%20term`:

The page's url is `..../sitepages/search.aspx#some term` (actually reached using a search box WP in another page).

Hardcoding `some term` in the query in the search result WP produces results, while dynamic query won't because of the encoding.

**To Reproduce**
In a modern page, put the two WPs:
* SearchBox WP:
  * connect its source to use the url query fragment (`#myterms`)
* Search results WP:
  * connect to SP Search
  * keep the default query `{searchterms}`
  * in the 3rd page use a dynamic value:
    * source : page environement
    * property : query string
    * query string: url fragment
(not sure of the english labels)

Save the page, and trigger a query with "some term"

**Expected behavior**
Spaces shouldn't be encoded and results should be returned.

**Screenshots**
image

Firefox

stevebeauge:
Attempt to fix #1056.

In my env, it behaves as expected

### DIFF
--- a/search-parts/src/webparts/searchResults/SearchResultsWebPart.ts
+++ b/search-parts/src/webparts/searchResults/SearchResultsWebPart.ts
@@ -254,7 +254,9 @@ export default class SearchResultsWebPart extends BaseWebPart<ISearchResultsWebP
         let renderRootElement: JSX.Element = null;
         let renderDataContainer: JSX.Element = null;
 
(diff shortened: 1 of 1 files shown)


wobba:
Thank you :)

Issue #725, opened 2021-01-31
v4.0 cannot find 'webpack' and other modules when setting up search-parts project


aliotsy:
**Version used**
Ex: 4.0

**Describe the bug**
While following the instructions for setting up the solution locally, 'webpack' module is not found when I run `gulp bundle` in the `search-parts` projects.

**To Reproduce**
Follow instructions at https://microsoft-search.github.io/pnp-modern-search/how-to-contribute/#setting-up-the-solution-locally
1. Open the `search-extensibility` project and run `npm i`, `gulp bundle`, and `npm link`.
2. Open the `search-parts` project and run npm link `@pnp/modern-search-extensibility` and `gulp bundle`.
3. See error

**Expected behavior**
`gulp bundle` runs, allowing me to run `npm run serve`.

**Desktop (please complete the following information):**
 - Browser Chrome, MacOS

**Additional context**
I'm able to address the error by running `npm link webpack` and then attempting gulp `bundle again`, but then encounter the same error with the following modules:

- @microsoft/sp-build-web
- webpack-bundle-analyzer
- colors
- @microsoft/sp-webpart-workbench/lib/api

Again, running `npm link` for each module seems to address the error, but then `gulp bundle` errors on `sass` because it can't find `node_modules/office-ui-fabric-react/dist/sass/_References.scss`. One more `npm link` for `office-ui-fabric-react` fixes that, but then gulp bundle gives me this:

![Screen Shot 2021-01-30 at 4 13 03 PM](https://user-images.githubusercontent.com/18564092/106371181-2f5d3500-6316-11eb-896b-650f9412cabc.png)

Gave up at that point.

FranckyC:
Hi @aliotsy , did you run `npm i ` in the `search-parts` project?

aliotsy:
> Hi @aliotsy , did you run `npm i ` in the `search-parts` project?

I did run `npm i` before `npm link`, and that didn't work, but just now I ran `npm i` after `npm link` but before `gulp bundle`, and that DID work! Thanks!

Might be nice to clarify that in step 2 of the documentation.

prabhu8323:
Hi @aliotsy , I have also faced the same issue as you. I have followed the steps as you did but still getting the error as in below screen shot:

![image](https://user-images.githubusercontent.com/48989167/107066202-ae190d00-6803-11eb-9c77-bc45d5f2e341.png)

aliotsy:
[...]

PR #1078, opened 2021-06-15
fix(doc): update contribution documentation

Laul0:
Take into account the @wobba comment https://github.com/microsoft-search/pnp-modern-search/issues/725#issuecomment-861607406.

The `@pnp/modern-search-extensibility` is an npm reference into the `search-parts` and `search-extensibility-demo` projects. Not no need to link the project locally except if some changes are made to it.

But, be aware about any change on `@pnp/modern-search-extensibility` is critical.


Laul0:
@wobba, it is better like this?

PR #1111, opened 2021-06-28
New scenarios

PerOve:
3 new scenarios.

Now with lowercase letters in all filenames. :)


PerOve:
@wobba I changed the filenames to lower case. That was not a fun process.
If I only knew that changing between lower cases and uppercases in windows is not registrered as a change in git...

Anyway, here are some new articles. Hope they are ok.

PerOve:
When will develop branch merge into main?

Maybe docs should have its own branch that can be merged into main more often? Since there is no breaking code? 🙂

wobba:
@PerOve I'm considering the best approach. The gh-pages branch pulls from main. We could merge to main every so often as well, which shouldn't be an issue as we have release tags if you want the release main code.

PerOve:
I would like to try the approach you land on.
Just let me know what I need to do, and I will write a new scenario and test.

🙂

wobba:
Create a PR and target main. I'll then just merge down to develop. That should work. I just published the recent docs as well.

PR #1115, opened 2021-06-28
fix(searchBox): search terms URL encode

### LINKED ISSUE #1096: Ampersand "&" in the search query is causing issue
My search box webpart is configured to refresh the URL parameters of the page environment when user keys in search query & hits on the search button.

Issue: However, when the user searches for any search query consisting of the "Ampersand" is resulting in inaccurate search results as the Search Results webpart is not omitting any text after the ampersand symbol.

E.g.: I enter "Test_value1 & Test_value2" in the search box & hit on search, it only searches for "Test_value1 " and omits the rest of the value. I tried with URL encoding of the ampersand, it didn't work either.  Please help me with this.

Thanks in advance.

Laul0:
This PR fix #1096

### DIFF
--- a/search-parts/src/helpers/UrlHelper.ts
+++ b/search-parts/src/helpers/UrlHelper.ts
@@ -64,10 +64,10 @@ export class UrlHelper {
             // Append new param
             const hasQuestionMark = /\?/.test(url);
             delimiter = hasQuestionMark ? "&" : "?";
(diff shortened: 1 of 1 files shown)


FranckyC:
Thanks @Laul0, issue confirmed and fix tested on your side?

Laul0:
Yes

Issue #1085, opened 2021-06-18
SearchRefiners: Unwanted filters checked when using filters in URL


salascz:
**Version used**
3.20

**Describe the bug**
We are calling the Search Result page with query filters defined in URL. (e.g. `<site>/SitePages/Results.aspx?filters=[{"n":"RefinableString151","t":["Getting%20Started"]}]` )

The results are returned correctly, but the Search Refiners WebPart select more checkboxes, than we expected.

I did a quick issue analyse:
- I have (probably) found the issue in SearchResultWebPart._mapDefaultSelectedFiltersToRefinementResults method. 
  - The "unwanted" filter is selected based on the last condition (`else if (StringHelper.longestCommonSubstring(updatedSelectedFilterValue.RefinementToken, refinementValue.RefinementValue)`)
  - Please, check our data (n = value from updatedSelectedFilter, r = matchingRefinementResults):
![image](https://user-images.githubusercontent.com/5662939/122541730-46cd5e00-d02a-11eb-8613-8db0351d1217.png)
  - the StringHelper is probably called with parameters `("ǂǂ47657474696e672053746172746564", "Beginner")` which return `"e"` as a result, condition is evaluated as a `true` and refinement is added to `updatedSelectedFilters` collection.

Not sure, if I understand the purpose of this condition, but it looks it can break the functionality in some cases...

Hope I was able to explain the problem. The "unwanted" filter was not selected anymore when I disabled this condition.

Thanks

wobba:
Fixed in the next release.

PR #1120, opened 2021-06-30
SearchRefiners: Unwanted filters checked when using filters in URL #1085

cebud:
SearchRefiners: Unwanted filters checked when using filters in URL #1085

If RefinementToken doesn't contains ǂǂ, then it could means that the provided condition in URL is an FQL expression


wobba:
Thank you :)

PR #1138, opened 2021-07-07
Fix for issue 1121

rdicicco:
Update SearchService.ts to check for valid token before trying to retrieve file  url


wobba:
Thank you

PR #1183, opened 2021-07-29
added ImportExport nl-nl translation

dennisvanniel:
Added nl-nl translation for key _'ImportExport'_ as this key was missing from nl-nl.js in webpart translation files.

tested by

- Adding updated solution package to dev tenant
- Adding the PnP webparts to page on sitecollection with nl-nl language setting
- Checking if the texts displayed in the webpart UI
- Checking if the webparts still function, separately and in coorperation
- finally checking the same points for the the webpart on en-US site


wobba:
Thank you :)

PR #1199, opened 2021-08-10
Use input query text Hover Message update

CreativeAcer:
`{searchQueryText}` no longer works but is still used in the hover message as a tip.  

`Use the {searchQueryText} token in your data source to retrieve this value`  

new value that does work is `{inputQueryText}`  
  
Example:  
`Path:https://tenant.sharepoint.com/sites/sitename/Lists/listname/* AND ListItemID:{inputQueryText}`

Files updated in the following location:  

- Search-parts/src/webparts/searchResults/loc/*


wobba:
Thank you :)

PR #1249, opened 2021-08-31
Fixed #1236: Double encoding for queryText in Search Box

### LINKED ISSUE #1236: Search Box Query String double encoding
**Version used**
4.3

**Describe the bug**

I'm using the Search Box Web Part to search for people in Microsoft Search. If I search by using a single word, e.g. Megan. Everything works. If I add any uriencodable characters such as a space, the search query parameter seems to be double uriencoded. See screenshots below.

**To Reproduce**

1. Install PnP Modern Search App to a site of your choice
2. Add a PnP Search Box Web Part to a page in the site.
3. Open Property pane for the Web Part
4. Set **Send the query to a new page** to On
5. Set **Page URL** to your tenants people search page _e.g. https://{mytenant}.sharepoint.com/_layouts/15/search.aspx/people_
6. Set **Method** to Query string parameter
7. Set **Parameter name** to q
8. Save your changes and publish the page
9. Try searching for e.g. **Megan Bowen**
10. You should navigate to the people search page, and see that the space in the query parameter is encoded twice as q=**Megan%2520Bowen**

**Expected behavior**
Query parameter should not be double encoded. Query parameter should be  **q=Megan%2520Bowen**

**Screenshots**

image
Example configuration

image
Search result page after searching for Megan Bowen

image
Query parameter double encoded

**Additional context**

I guess this is a similar case as in issue #20.

I think I found the culprit of the issue here

https://github.com/microsoft-search/pnp-modern-search/blob/b78aba6323a9d12ae2cca39f709f933d0f3f1746/search-parts/src/webparts/searchBox/components/SearchBoxContainer.tsx#L94

The querystring is encoded before the function call in line 94 and encoded again within the function.

If you take a look at the other use of the same function here, the string is not encoded before the function call.

https://github.com/microsoft-search/pnp-modern-search/blob/b78aba6323a9d12ae2cca39f709f933d0f3f1746/search-parts/src/webparts/searchBox/SearchBoxWebPart.ts#L651

jhholm:
Changed line 93 on SearchBoxContainer.tsx to use queryText instead of urlEncodedQueryText that caused double encoding on queryText.

### DIFF
--- a/search-parts/src/webparts/searchBox/components/SearchBoxContainer.tsx
+++ b/search-parts/src/webparts/searchBox/components/SearchBoxContainer.tsx
@@ -90,7 +90,7 @@ export default class SearchBoxContainer extends React.Component<ISearchBoxContai
           newUrl = searchUrl.href;
         }
         else {
-          newUrl = UrlHelper.addOrReplaceQueryStringParam(searchUrl.href, this.props.queryStringParameter, urlEncodedQueryText);
+          newUrl = UrlHelper.addOrReplaceQueryStringParam(searchUrl.href, this.props.queryStringParameter, queryText);
         }
 
         // Send the query to the new page


wobba:
Have you tested it on-page as well? If you set to use query param, and that when you search it updates the URL param correct.. and it works on refresh?

jhholm:
The code path only applies when sending the search results to a new page:
https://github.com/microsoft-search/pnp-modern-search/blob/b78aba6323a9d12ae2cca39f709f933d0f3f1746/search-parts/src/webparts/searchBox/components/SearchBoxContainer.tsx#L79

It also only applies when you are excplicitly using query parameters (the else statement from this line)
https://github.com/microsoft-search/pnp-modern-search/blob/b78aba6323a9d12ae2cca39f709f933d0f3f1746/search-parts/src/webparts/searchBox/components/SearchBoxContainer.tsx#L89

I did also test how Search Box itself parses a query parameter from the page:
image

In this case the parameter can be both Megan%20Bowen or MEgan%2520Bowen, so that's interesting. Though I did not touch that part of the code at all.

wobba:
Thank you :)

Issue #1251, opened 2021-08-31
Replace default image for the folder view


sb-sheriff:
Hi,

Did anyone get experience to replace the broken default image as shown below?

![image](https://user-images.githubusercontent.com/87348254/131589298-1c1bd02b-7bda-4157-9c26-cd5fbdb590ef.png)

FranckyC:
Hi @sb-sheriff, this is not a broken image, just a placeholder image when no preview image is available for the result. In your example, you are displaying folders so it is the expected behavior.

FranckyC:
If you want to get rid of if, simply update the Handlebars template logic.

sb-sheriff:
I see. Yes, I'm looking to replace this placeholder image with an image of the regular folder.  May I get to know where to find the Handlebars template logic?

wobba:
@sb-sheriff You can use the result type option for example - and provide a custom template for folders. I'll tag this as an enhancement to skip preview for folders, as it doesn't make sense to have it.

sb-sheriff:
@wobba OK, now I need to find out where to use the result type option/custom template -- this one? 
https://github.com/microsoft-search/pnp-modern-search/blob/main/docs/scenarios/edit-templates-using-vscode-and-onedrive.md

I thought I could simply replace that default image with a regular folder image and not confuse the user as something broken.

wobba:
I assume you're using v4. Take a look at https://microsoft-search.github.io/pnp-modern-search/usage/search-results/layouts/#result-types

wobba:
Added a fix for this and also made sure OneNote notebooks work as expected as they had folder icons.

sb-sheriff:
Copy that, thanks!

wobba:
Re-open until we get the change into v4 as well.

wobba:
Fixed for the SP provider in v4. Awaiting the Graph provider which seems to have API issues.

sb-sheriff:
> Fixed for the SP provider in v4. Awaiting the Graph provider which seems to have API issues.

Supposed I could download and reinstall the v4 to fix the issue? ... if the Graphy provider resolved the API issue...

PR #1277, opened 2021-09-08
Fix preview for folders and fix icon for onenote's

wobba:
#1251


wobba:
Haven't been able to verify for Graph provider as properties are not returned.

PR #1282, opened 2021-09-10
Adds ability to sort columns with handlebar expressions on Details List

ellinge:
Since there's a check if a column contains a handlebars expression which disallows sorting I've added the possibility to reference a sorting field when handlebars and sorting is checked.

The settings:
bild

And the result:
bild


ellinge:
Sonar is regarding the language files/resources as code duplication

PzKfWg:
Good job! I need exactly this :-)

search-parts/src/components/DetailsListComponent.tsx
```diff
@@ -228,17 +233,17 @@ export class DetailsListComponent extends React.Component<IDetailsListComponentP
...
             sortDescendingAriaLabel: 'Sorted Z to A',
-            onColumnClick: column.enableSorting && !column.useHandlebarsExpr ? this._onColumnClick : null,
+            onColumnClick: allowSorting ? this._onColumnClick : null,
             data: {
               // Set arbitrary data for the current column
               useHandlebarsExpr: column.useHandlebarsExpr,
 
               // Set the column value for sorting (i.e field to use)
-              value: column.value
+              value: (column.useHandlebarsExpr && column.valueSorting) || column.value
```
FranckyC, line 246: Should be more `value: column.valueSorting ? column.valueSorting : column.value` meaning if a alternative sort field is set, the column value is always an Handlebars expression based on column possibilities (see comment below).

search-parts/src/components/DetailsListComponent.tsx
```diff
@@ -217,7 +222,7 @@ export class DetailsListComponent extends React.Component<IDetailsListComponentP
     if (this.props.columnsConfiguration) {
 
       this.props.columnsConfiguration.forEach((column: IDetailsListColumnConfiguration) => {
-
+        let allowSorting = column.enableSorting && (column.useHandlebarsExpr !== true || (column.valueSorting && column.valueSorting.length > 0));
```
FranckyC, line 225: Leaving just `column.enableSorting` should be sufficient now as we allow sort with both value/Handlebars expression now. And should be a 'const' here.

search-parts/src/layouts/results/detailsList/DetailListLayout.ts
```diff
@@ -162,6 +162,23 @@ export class DetailsListLayout extends BaseLayout<IDetailsListLayoutProperties>
                             );
                         }
                     },
+                    {
```
FranckyC, line 165: The 'valueSorting' column should only appear when Handlebars expression and 'sortable' options are set together. Otherwise, this column should be hidden and **empty**:

image

Moreover, in the case the user revert back to a simple value, the `valueSorting` value should be reset every time.
Last thing, if the 'sortable' option is set but no value is set for 'valueSorting' then the configuration should not be saved (e.g. required field in this specific).

search-parts/src/layouts/results/detailsList/DetailListLayout.ts
```diff
@@ -162,6 +162,23 @@ export class DetailsListLayout extends BaseLayout<IDetailsListLayoutProperties>
...
+                        id: 'valueSorting',
+                        title:strings.Layouts.DetailsList.ValueSortingColumnLabel,
+                        type: this._customCollectionFieldType.custom,
+                        onCustomRender: (field, value, onUpdate, item, itemId, onCustomFieldValidation) => {
```
FranckyC, line 169: The control used here shouldn't be a `TemplateValueFieldEditor` but an `AsyncCombo`. See other examples in the code (`React.createElement(AsyncCombo, {...`

FranckyC (changes requested):

search-parts/src/components/DetailsListComponent.tsx
ellinge: This will not be backwards compatible since it will try to start sorting existing webparts with enableSorting set on handlebars expressed fields since that's a possibility (but it just does not work).

search-parts/src/layouts/results/detailsList/DetailListLayout.ts
ellinge: https://pnp.github.io/sp-dev-fx-property-controls/controls/PropertyFieldCollectionData I can't find any way to update the field collection depending on the current state of all the items (any, sortable + handlebars). But setting the column label to '' sorted it out.
image
image
Is this acceptable?

ellinge: There does not seem any other way to hide the column then setting lable to ''. This is also used for onCustomFieldValidation so I placed the requirement on enableSorting instead to be able to fulfill both these wishes.

search-parts/src/layouts/results/detailsList/DetailListLayout.ts
ellinge: Changed to AsyncCombo now.

search-parts/src/components/DetailsListComponent.tsx
ellinge: Since we added the validation for when valueSorting is possible (and cleared) there should be no need to check handleBarsExps here.

ellinge:
Finally some additional notes/findings:
- Could not find any other event which checked the current state of row / all rows to add conditional checks for header or validation while editing. onPropertyUpdate gets called after save and it then the propertyFieldCollection gets reinitialized again. I had to render custom checkboxes to do these checks. Tried getting [onGetErrorMessage ](https://github.com/pnp/sp-dev-fx-property-controls/blob/7fb3a42fcf61e69c49b7ba589dfa941b7ada62cd/src/propertyFields/collectionData/ICustomCollectionField.ts#L54)going which seems to enable conditional formatting on the current item but that did not trigger for me (it’s not triggered for custom or Boolean fields if checking code).
- https://github.com/pnp/sp-dev-fx-property-controls/blob/7fb3a42fcf61e69c49b7ba589dfa941b7ada62cd/src/loc/mystrings.d.ts does not seem to be exposed in the bundle for @pnp/spfx-property-controls so had to add typing for Required myself. Do we need to add a custom validation instead like for the label?

search-parts/src/layouts/results/detailsList/DetailListLayout.ts
FranckyC: > 
> 
> https://pnp.github.io/sp-dev-fx-property-controls/controls/PropertyFieldCollectionData I can't find any way to update the field collection depending on the current state of all the items (any, sortable + handlebars). But setting the column label to '' sorted it out.
> image
> image
> Is this acceptable?

Yes that would be a solution but it only works for the initial state unfortunately, not when you configure fields dynamically. Maybe you could just let the title always empty and add a label next to the combo box, like this (juste an example):

image

This way we get rid of the empty column.

FranckyC: > 
> 
> There does not seem any other way to hide the column then setting lable to ''. This is also used for onCustomFieldValidation so I placed the requirement on enableSorting instead to be able to fulfill both these wishes.

I'm fine with this.

FranckyC:
> 
> 
> Finally some additional notes/findings:
> 
>     * Could not find any other event which checked the current state of row / all rows to add conditional checks for header or validation while editing. onPropertyUpdate gets called after save and then the propertyFieldCollection gets reinitialized again. I had to render custom checkboxes to do these checks. Tried getting [onGetErrorMessage ](https://github.com/pnp/sp-dev-fx-property-controls/blob/7fb3a42fcf61e69c49b7ba589dfa941b7ada62cd/src/propertyFields/collectionData/ICustomCollectionField.ts#L54)going which seems to enable conditional formatting on the current item but that did not trigger for me (it’s not triggered for custom or Boolean fields if checking code).
> 
>     * https://github.com/pnp/sp-dev-fx-property-controls/blob/7fb3a42fcf61e69c49b7ba589dfa941b7ada62cd/src/loc/mystrings.d.ts does not seem to be exposed in the bundle for @pnp/spfx-property-controls so had to add typing for Required myself. Do we need to add a custom validation message instead like for the label?

Except the column header, I'm ok with the proposed solution. If you can make the requested update regarding the label next to the combo box, we are good to go I think. Nice job!

ellinge:
Regarding the column header. The default values causes it to be visible. The label is also used as a placeholder text. Wouldn’t that suffice? 

At tops removing the conditional header.

FranckyC:
> 
> 
> Regarding the column header. The default values causes it to be visible. The label is also used as a placeholder text. Wouldn’t that suffice?
> 
> At tops removing the conditional header.

Yes basically the idea is to remove the header title to make the column "hidden" in any case. You can still use the label as placeholder text.

ellinge:
https://github.com/microsoft-search/pnp-modern-search/blob/b73293d9dbec3c1c05c4321a5ba01a962ffc8941/search-parts/src/layouts/results/detailsList/DetailListLayout.ts#L180 This placeholder should be enough then?

FranckyC:
> 
> 
> https://github.com/microsoft-search/pnp-modern-search/blob/b73293d9dbec3c1c05c4321a5ba01a962ffc8941/search-parts/src/layouts/results/detailsList/DetailListLayout.ts#L180
> This placeholder should be enough then? Both feels a bit superfluous and different compared to the other columns, if not in the header.

Just add a label next to the combo please. I've just tested and the placeholder is not visible for default values.

FranckyC:
Thanks @ellinge

PR #1292, opened 2021-09-14
Synonym Feature based on a SharePoint List

Laetterman:
Adds the synonym functionality to the search where the synonyms are managed in an external SharePoint list.


Laetterman:
Hi
Why does the quality check fail? I do not see what I have done wrong...

search-parts/config/package-solution.json
```diff
@@ -3,7 +3,7 @@
     "solution": {
         "name": "PnP Modern Search - Search Web Parts - v4",
         "id": "59903278-dd5d-4e9e-bef6-562aae716b8b",
-        "version": "4.3.0.0",
+        "version": "4.3.0.2",
```
wobba, line 6: We'll bump the version number :)

search-parts/src/dataSources/MicrosoftSearchDataSource.ts
```diff
@@ -161,6 +212,17 @@ export class MicrosoftSearchDataSource extends BaseDataSource<IMicrosoftSearchDa
         };
 
         const searchRequest = await this.buildMicrosoftSearchRequest(dataContext);
+
+        // enrich the query with synonyms if enabled....
+        if (this.properties.synonymsEnabled) {
+            if (this._synonymsList == undefined) {
+                // if the synonyms list has not been loaded during startup/initalization (typically ind debug/dev mode) load it again here
+                console.log("initializing/loading the synonyms table at query time...");
```
wobba, line 220: Please use a logging provider and not console.log

wobba:
Before we review... you more or less moved over the v3 approach right?

search-parts/src/dataSources/MicrosoftSearchDataSource.ts
```diff
@@ -109,9 +146,16 @@ export class MicrosoftSearchDataSource extends BaseDataSource<IMicrosoftSearchDa
...
 
+    // for synonyms functionality
+    private _synonymsService: ISynonymsService;
+    private _synonymsList: ISynonymsListEntry[];
+
     public constructor(serviceScope: ServiceScope) {
         super(serviceScope);
 
+        // for synonyms functionality
+        this._synonymsService = new SynonymsService();
```
FranckyC, line 157: Please use service scopes to instanciate service instances (See other services)

search-parts/src/dataSources/SharePointSearchDataSource.ts
```diff
@@ -108,6 +114,38 @@ export interface ISharePointSearchDataSourceProperties {
...
+
+    /**
+     * Name of the synonyms column (text type field)
+     */
+     synonymsFieldNameSynonyms : string;
+
+    /**
+     * Name of the mutual flag column (boolean type)
+     */
+     synonymsFieldNameMutual : string;
```
FranckyC, line 147: Please use a single interface to group all these properties (like ISynonymsProps)

search-parts/src/dataSources/SharePointSearchDataSource.ts
```diff
@@ -140,9 +178,16 @@ export class SharePointSearchDataSource extends BaseDataSource<ISharePointSearch
...
 
+    // for synonyms functionality
+    private _synonymsService: ISynonymsService;
+    private _synonymsList: ISynonymsListEntry[];
+
     public constructor(serviceScope: ServiceScope) {
         super(serviceScope);
 
+        // for synonyms functionality
+        this._synonymsService = new SynonymsService();
```
FranckyC, line 189: Same comment as above

search-parts/src/dataSources/SharePointSearchDataSource.ts
```diff
@@ -188,11 +233,29 @@ export class SharePointSearchDataSource extends BaseDataSource<ISharePointSearch
...
     }
 
     public async getData(dataContext: IDataContext): Promise<IDataSourceData> {
 
         const searchQuery = await this.buildSharePointSearchQuery(dataContext);
+
+        // enrich the query test with synonyms (if enabled)...
+        // we do it here and not in the SharePoint Search Service as for MS Search there is no Search Service
+        if (this.properties.synonymsEnabled) {
+            if (this._synonymsList == undefined) {
```
FranckyC, line 251: Use strict equality "==="

FranckyC (changes requested):
Hi @Laetterman, please first fix these issues before merging.

search-parts/config/package-solution.json
Laetterman: Absolutely fine. It was only for our internal purposes to have different versions than the main version.
Is it possible to exclude this file from the code review?

Laetterman:
Hi
Why does the quality check fail? I do not see what I have done wrong... :-)

search-parts/src/dataSources/MicrosoftSearchDataSource.ts
Laetterman: This is no real logging. Just an easy way out to see when the synonym table gets loaded... :-) I'll remove the output.

Laetterman:
> Before we review... you more or less moved over the v3 approach right?

Yes. It's a new approach. One point ist that the synonyms are managed over a SharePoint list and not in the webpart itself. The main reason was that super users can manage the synonyms without having to update the webpage/page. The second reason was that (in our case) the List gets poartially updated from an external app.

The second main change is the way of lookup: in v3 the synonym entries where compared with the query. In our approach we use the oposite way: we analyze the query and use the query terms/combinations to look up for synonyms.

This we have implemented by the request of the customer and we thaugt it makes sense to make it available for others also... :-)

Regards
Marc

search-parts/src/dataSources/MicrosoftSearchDataSource.ts
Laetterman: Thanks for the review Franck. We will have a look at all your points you mantioned. :-)
Marc

search-parts/config/package-solution.json
Laetterman: I have reverted it in the latest commit

search-parts/src/dataSources/MicrosoftSearchDataSource.ts
Laetterman: Fixed in tle latest commit

search-parts/src/dataSources/MicrosoftSearchDataSource.ts
Laetterman: Fixed in the latest commits

search-parts/src/dataSources/SharePointSearchDataSource.ts
Laetterman: Fixed/Changed in the latest commit

search-parts/src/dataSources/SharePointSearchDataSource.ts
Laetterman: Fixed in the latest commit

search-parts/src/dataSources/SharePointSearchDataSource.ts
Laetterman: Done in the latest commit

dfrancoeur:
@Laetterman Just curious about your requirement. Is the idea that the client would have a master translation table for their whole environment, stored in this list? I've seen this approach before (LiveTiles for example does this) but I worried about long term scalability, especially when the term store already has a translation feature. Do you think it would be possible to build on the general approach to lookup translations in the term store, and then append/modify the search query to look for those alternate values more dynamically?

Laetterman:
> @Laetterman Just curious about your requirement. Is the idea that the client would have a master translation table for their whole environment, stored in this list? I've seen this approach before (LiveTiles for example does this) but I worried about long term scalability, especially when the term store already has a translation feature. Do you think it would be possible to build on the general approach to lookup translations in the term store, and then append/modify the search query to look for those alternate values more dynamically?

The requirement has nothing to do with translations. Where do you get this information? It is the synonym feature introduced in v3 just with the oposite lookup aproach and management in a sharepoint list. :-)

dfrancoeur:
> > @Laetterman Just curious about your requirement. Is the idea that the client would have a master translation table for their whole environment, stored in this list? I've seen this approach before (LiveTiles for example does this) but I worried about long term scalability, especially when the term store already has a translation feature. Do you think it would be possible to build on the general approach to lookup translations in the term store, and then append/modify the search query to look for those alternate values more dynamically?
> 
> The requirement has nothing to do with translations. Where do you get this information? It is the synonym feature introduced in v3 just with the oposite lookup aproach and management in a sharepoint list. :-)

Got it. My head just went there due to my own bias. Regardless I have the same question about the Term Store and the existing Synonyms. I imagine the client just preferred the list?

FranckyC:
@Laetterman thanks for the updates
@wobba please do not merge yet. Not tested yet.

Laetterman:
> > > @Laetterman Just curious about your requirement. Is the idea that the client would have a master translation table for their whole environment, stored in this list? I've seen this approach before (LiveTiles for example does this) but I worried about long term scalability, especially when the term store already has a translation feature. Do you think it would be possible to build on the general approach to lookup translations in the term store, and then append/modify the search query to look for those alternate values more dynamically?
> > 
> > 
> > The requirement has nothing to do with translations. Where do you get this information? It is the synonym feature introduced in v3 just with the oposite lookup aproach and management in a sharepoint list. :-)
> 
> Got it. My head just went there due to my own bias. Regardless I have the same question about the Term Store and the existing Synonyms. I imagine the client just preferred the list?

We also discussed the management the synonyms in the Termstore. But we / the customer found that the Termstore management for 'simple synonym' management is not really comfortable, even in the new modern design and gaining Management access to a certain User Group is easier than to the termstore.... :-)

But I think that our integration is open/flexible so that the 'fill' of the (internal) synonyms list could also come from a termstore lookup instead of a sharepoint list...

dfrancoeur:
> > > > @Laetterman Just curious about your requirement. Is the idea that the client would have a master translation table for their whole environment, stored in this list? I've seen this approach before (LiveTiles for example does this) but I worried about long term scalability, especially when the term store already has a translation feature. Do you think it would be possible to build on the general approach to lookup translations in the term store, and then append/modify the search query to look for those alternate values more dynamically?
> > > 
> > > 
> > > The requirement has nothing to do with translations. Where do you get this information? It is the synonym feature introduced in v3 just with the oposite lookup aproach and management in a sharepoint list. :-)
> > 
> > 
> > Got it. My head just went there due to my own bias. Regardless I have the same question about the Term Store and the existing Synonyms. I imagine the client just preferred the list?
> 
> We also discussed the management the synonyms in the Termstore. But we / the customer found that the Termstore management for 'simple synonym' management is not really comfortable, even in the new modern design and gaining Management access to a certain User Group is easier than to the termstore.... :-)
> 
> But I think that our aproach is open/flexible so that the 'fill' of the (internal) synonyms list could also come from a termstore lookup instead of a sharepoint list...

I get it, thanks for confirming and appreciate the guidance. Our team will likely take a look at this.

Laetterman:
> @Laetterman thanks for the updates
> @wobba please do not merge yet. Not tested yet.

Thanks Franck and wobba. Sure, i also apreciate testing from your side and not 'blindly merge'! :-)
Greetings from Switzerland
Martc

Laetterman:
Hi @wobba 
What are the next steps? How can I resolve the 'SonarCloud Quality Gate failed' issue? I have just added new translations and do not see what I have done wrong...
Regards
Marc

wobba:
You can likely ignore sonarcloud. Still haven't had time to test this one.

FranckyC:
Hi @Laetterman, didn't have time to test. Did you update the **SynonymsService** to not use @pnp library?

Laetterman:
Hi @FranckyC 

Not sure what you really mean... :-)

Regards
Marc

> Hi @Laetterman, didn't have time to test. Did you update the **SynonymsService** to not use @pnp library?

FranckyC:
> 
> 
> Hi @FranckyC
> 
> Not sure what you really mean... :-)
> 
> We are using the whree pnp references/imports (web, list, items) to access/load the SP List which contains the Synonyms. Is there something wrong with that?
> 
> image
> 
> Regards
> Marc
> 
> > Hi @Laetterman, didn't have time to test. Did you update the **SynonymsService** to not use @pnp library?

Yes. Services in the solution don't use the pnp library as it increases the bundle size for nothing. You can easily achieve the same thing using the `SPHttpClient` class.

Laetterman:
> > Hi @FranckyC
> > Not sure what you really mean... :-)
> > We are using the whree pnp references/imports (web, list, items) to access/load the SP List which contains the Synonyms. Is there something wrong with that?
> > image
> > Regards
> > Marc
> > > Hi @Laetterman, didn't have time to test. Did you update the **SynonymsService** to not use @pnp library?
> 
> Yes. Services in the solution don't use the pnp library as it increases the bundle size for nothing. You can easily achieve the same thing using the `SPHttpClient` class.

@FranckyC Ahh, ok. We will have a look at it.

wobba:
One option is to have pnpjs as part of a separate bundle/chunk when you enable the synonym feature. Loading extra bytes for this feature is ok imo. I'll however let @FranckyC decide on this one.

FranckyC:
> 
> 
> One option is to have pnpjs as part of a separate bundle/chunk when you enable the synonym feature. Loading extra bytes for this feature is ok imo. I'll however let @FranckyC decide on this one.

Absolutely no value to add this library here so we stick to the `SPHttpClient`. Read items in a list shouldn't require a third party honestly...

docs/usage/search-results/common/synonyms.md
```diff
@@ -0,0 +1,35 @@
+Synonyms functionality/configuration is available for both SharePoint Search and Microsoft Search data sources.
+
+If this function is enabled, the query terms are expanded with synonyms configured in an external SharePoint list.
+Depending on a 'mutual' flag the synonyms are expanded in one direction only (keyword --> synonyms) or mutual (keywords <--> synonyms <--> synonyms).
```
wobba, line 4: Better to name it two-way than mutual in my opinion.

search-parts/src/loc/da-dk.js
```diff
@@ -305,6 +305,15 @@ define([], function() {
...
                 }
+            },
+            Synonyms: {
+                SynonymSettingsGroupName: "Synonymer",
+                EnableSwitchLabel: "Aktivere Synonymer",
+                SiteUrlLabel: "Webstedets URL for synonymlisten",
+                SynonymListLabel: "Synonymer Liste Navn",
+                SynonymListFieldNameKeyword: "Nøgleord Feltnavn (normalt 'title' feltet)",
+                SynonymListFieldNameSynonyms: "Synonyms Feltnavn (tekst)",
+                SynonymListFieldNameMutual: "Gensidigt feltnavn (boolean)"
```
wobba, line 316: Boolean fields in SharePoint are name Yes/No

search-parts/src/services/synonymsService/SynonymsService.ts
```diff
@@ -0,0 +1,164 @@
...
+        // initalizing an empty return value...
+        let synonyms: ISynonymsListEntry[] = [];
+
+        try {
+            if (webUrl == "" || listName == "" || fieldNameKeyword == "" || fieldNameSynonyms == "" || fieldNameMutual == "") {
+                throw new Exception("Synonyms configuration not completed or wrong");
+            }
+            const web = Web(webUrl);
+            const response = await web();
+            const items: any[] = await web.lists.getByTitle(listName).items.get();
```
wobba, line 153: Some caching here would be great as reading this list upon every page load by ever user can cause unnecessary API requests. Maybe introduce a setting to set the cache interval where default is X days.

wobba:
@Laetterman These don't expand if `foo` is defined as a synonym:

`(foo)`
`(foo OR asd)`
`ANY(foo)` <- I'm ok with ANY/WORDS not expanding, as they are OR/AND inside the operator
`-foo`
`foo foo` <- turns into `(foo foo) OR (bar bar) OR (bar bar)` which might be ok, or should one permutation be "foo bar"?

docs/usage/search-results/common/synonyms.md
Laetterman: We named it mutual by design, because if we have multiple synonyms for a keyword then 2way in our opinion means only the beack combination between the found synonym and the keyword and not amons all synonyms. :-)

Laetterman:
> > One option is to have pnpjs as part of a separate bundle/chunk when you enable the synonym feature. Loading extra bytes for this feature is ok imo. I'll however let @FranckyC decide on this one.
> 
> Absolutely no value to add this library here so we stick to the `SPHttpClient`. Read items in a list shouldn't require a third party honestly...

@wobba @FranckyC We will change it to the SPHttpClient

Laetterman:
> @Laetterman These don't expand if `foo` is defined as a synonym:
> 
> `(foo)`
> `(foo OR asd)`
> `ANY(foo)` <- I'm ok with ANY/WORDS not expanding, as they are OR/AND inside the operator
> `-foo`
> `foo foo` <- turns into `(foo foo) OR (bar bar) OR (bar bar)` which might be ok, or should one permutation be "foo bar"?

@wobba the 'ANY' operator iscurrently not stripped from the query term to get the expansion (only AND, OR and NOT as in the v3 Version). The same for the - operator. We will have a look at those special combinations...
But for the - Operateor we have to check with the customer if an exclusion should also exclude synonyms. In my opinion it should not because we are then hiding documents which the user does not understand...

For the 'optimization' of the final query: we had not the intend to re-analyze and optimize the final query as this is a compley thing and is done by the query-Analyzer in the backend. We wanted to have a syntax which is 'readable' if you have to analyze it with the dabugger. :-)  
--> Normally the end users in the UI do 98% not enter that complex queries so that the effort would be too big for a heavy optimizatuion and special cases like 'foo foo' example; the result has to be ok... :-)

search-parts/src/services/synonymsService/SynonymsService.ts
Laetterman: We will have a look at it what the options are here.

Normlly the Page is loaded once and not between multiple queries. Therefore we decided to load the list at page load (as fallback on first query, eg. in debug mode). --> The list is only loaded once at page load and cached between queries
Anpother thing is, that it was a customer request to have synonyms active immediately.

search-parts/src/loc/da-dk.js
Laetterman: I will rename it

search-parts/src/services/synonymsService/SynonymsService.ts
wobba: @Laetterman Which is why a cache timeout is useful. Set to 0 for no cache, or maybe have hours/minutes as the other option.

If someone adds synonyms to a web part on an intranet home page with 100,000 employees you will have a lot of API calls not needed ;)

And, immediate synonyms is typically important for those testing, not the average person - as they have lived without the synonym up until it was added anyways 😉

Laetterman: @wobba Makes absolutely sense (was/is not an issue in our project, because the Webparts are used at a less frequently used location :-)

In terms of caching: did you meant using cookies or do you have any other mechanism in mind?

wobba:
I know what I list are edge cases which are typically not hit - but wanted to highlight them. You can address limitation in the docs as well. Can the docs state if column name should be internal name or column title names? Maybe add a comment to set the synonym table not to be indexed as well.

Laetterman:
> I know what I list are edge cases which are typically not hit - but wanted to highlight them. You can address limitation in the docs as well. Can the docs state if column name should be internal name or column title names? Maybe add a comment to set the synonym table not to be indexed as well.

Hi @wobba : absolutely fine with that. :-) I will update the docs with your hint. Especially excluding from index is a good point (as we created the list by the customers deployment script we have it already excluded over the script :-) but for manual cases you propably forget... Thanks Mikael

Info: we are currently working also on the other topics you and Franck mentioned. But as we are on heavy load at the moment it will take a little time... :-)

Laetterman:
@wobba @FranckyC 

Hi Mikael and Franck
Could you lready have a look at the changes? What are the next steps? :-)
Regards
Marc

wobba:
We will assess this one for the next release, so be patient :)

Laetterman:
> We will assess this one for the next release, so be patient :)

No worries Mikael. Just was not sure if my last checkin/update was technically correct as there are a lot of comments and changes to the dev branch meanwhile... :-)

wobba:
I've been thinking. Would it be possible to create this as a search query extension (https://docs.microsoft.com/en-us/sharepoint/dev/spfx/building-search-extensions#associating-query-extensions-to-sites). The setup would be a bit more tricky though. 

Or a separate non-rendering web part exposed as a data source. This would allow it to be more decoupled. Connect the result webpart to the synonym part, and then the synonym part to the search box or query param etc.

Nothing definitive here and I like to have the capability in - just exploring options to decouple things.

Laetterman:
> I've been thinking. Would it be possible to create this as a search query extension (https://docs.microsoft.com/en-us/sharepoint/dev/spfx/building-search-extensions#associating-query-extensions-to-sites). The setup would be a bit more tricky though.
> 
> Or a separate non-rendering web part exposed as a data source. This would allow it to be more decoupled. Connect the result webpart to the synonym part, and then the synonym part to the search box or query param etc.
> 
> Nothing definitive here and I like to have the capability in - just exploring options to decouple things.

Hi Mikael

Sure, we can think about it. :-)

We have built up the Feature on Customer Request (Nello from Mobiliar, I think you know him) as it is needed for their Project. We it object structally as it is for the taxonomy service or tocen service to decouple it as far as possible. Also we tried to integrate it in a way that it does not have a negative impact on the main functionality in case of failures... :-)

Unfortunately we do not have the ressources and time (and money ;-) ) to do a complete refactoring in this project and timeline... Therefore, we would appreciate if we could leave it as it is for the moment and maybe refactor it in a future version.  

Best regards
Marc

wobba:
@FranckyC what do you think... what's the effort needed to potentially remake this as a separate part to connect on the query chain, or should we just merge as is? I think I'm ok with this implementation.

wobba:
I'm changing this to draft to see if we can isolate it as an extension you can install alongside the core package and have it work. Converting to draft for now to have the code as reference.

Laetterman:
Hi @wobba @dfrancoeur 

How does it look like now? The go live date is comming closer now and we would like to use the official branch. Will this feature be merged now or do we have to manage our separate repository/branch?

I do not see where the problem is as it is a proper implementation as other services, properly documented and also a completely independent encapsulated feature...

Regards
Marc

wobba:
Hi, sorry for not following up. I still want to create this as an extension which can be dynamically added if you need it. So I'll keep the PR for code reference. For now you will have to have a custom branch.

edarroudi:
@wobba as we have the query modifiers now I implemented this using query modifiers with minor changes and it works so far.
I'm thinking of adding it to the samples repository and creating a scenario documentation sometime in the future. 
I want to think additionally about some changes  like 

1. paging due to thresholds for query item count as this would not work with many items 
2. directly loading a JSON file.
3. Other controls / Options

I just have a problem in understanding why the whole thing is built using **"OR"** statements instead of using the **"WORDS"** statement as that actually is called a [SYNONYM operator](https://learn.microsoft.com/en-us/sharepoint/dev/general-development/keyword-query-language-kql-syntax-reference#synonym-operators)

Is there a specific reason? Is this due to the fact that WORDS can not be used in property restrictions and OR can be used there?

Would be happy on any feedback on this! Thank you.

LeonArmston:
> @wobba as we have the query modifiers now I implemented this using query modifiers with minor changes and it works so far. I'm thinking of adding it to the samples repository and creating a scenario documentation sometime in the future. I want to think additionally about some changes like
> 
> 1. paging due to thresholds for query item count as this would not work with many items
> 2. directly loading a JSON file.
> 3. Other controls / Options
> 
> I just have a problem in understanding why the whole thing is built using **"OR"** statements instead of using the **"WORDS"** statement as that actually is called a [SYNONYM operator](https://learn.microsoft.com/en-us/sharepoint/dev/general-development/keyword-query-language-kql-syntax-reference#synonym-operators)
> 
> Is there a specific reason? Is this due to the fact that WORDS can not be used in property restrictions and OR can be used there?
> 
> Would be happy on any feedback on this! Thank you.

Hi @edarroudi 

Looks like you made awesome progress with this in December last year. Did you ever get close to adding it to the samples gallery and writing a scenario document up?

I have a current need for synonyms in search for a client and would love to know if you or anyone cracked it in PnP Modern Search v4?

I'm unable to code anything to create a pull request but very happy to test if you have anything that can be deployed and provide feedback.

Thanks

Leon

PR #1315, opened 2021-09-19
Feature/1188

wobba:


search-parts/src/webparts/searchResults/components/SearchResultsContainer.tsx
```diff
@@ -275,7 +275,7 @@ export default class SearchResultsContainer extends React.Component<ISearchResul
                 const hasContentClass = !isEmpty(contentClass);
                 const isLibItem = hasContentClass && contentClass.indexOf("Library") !== -1;
 
-                let contentTypeId = ObjectHelper.byPath(item, "ContentTypeId");
+                let contentTypeId = ObjectHelper.byPath(item, slots[BuiltinTemplateSlots.IsFolder]);
```
wobba, line 278: @FranckyC If I fix this, then click url's break for folders/sites/onenote. Which means it's an issue that we use AutoPreviewUrl for click and iframe preview. Which means the check should probably be removed here.

search-parts/src/webparts/searchResults/components/SearchResultsContainer.tsx
```diff
@@ -275,7 +275,7 @@ export default class SearchResultsContainer extends React.Component<ISearchResul
                 const hasContentClass = !isEmpty(contentClass);
                 const isLibItem = hasContentClass && contentClass.indexOf("Library") !== -1;
 
-                let contentTypeId = ObjectHelper.byPath(item, "ContentTypeId");
```
wobba, line 278: @FranckyC  If I fix this, then click url's break for folders/sites/onenote. Which means it's an issue that we use AutoPreviewUrl for click and iframe preview. Which means the check should probably be removed here in terms of click url logic.

FranckyC:
Thanks @wobba , I'll take a look

search-parts/src/webparts/searchResults/components/SearchResultsContainer.tsx
FranckyC: @wobba , we could simply change the slot used in templates from `PreviewUrl` to `Path` like in the first version to make a clear distinction.

search-parts/src/webparts/searchResults/components/SearchResultsContainer.tsx
```diff
@@ -303,7 +302,10 @@ export default class SearchResultsContainer extends React.Component<ISearchResul
                 let listId = ObjectHelper.byPath(item, slots[BuiltinTemplateSlots.ListId]);
                 let itemId = ObjectHelper.byPath(item, slots[BuiltinTemplateSlots.ItemId]); // Could be UniqueId or item ID
 
-                if (siteId && listId && itemId) {
+                let contentTypeId = ObjectHelper.byPath(item, slots[BuiltinTemplateSlots.IsFolder]);
+                const isContainer = contentTypeId ? contentTypeId.indexOf('0x0120') !== -1 : false;
```
FranckyC, line 306: This line is incorrect. If _IsFolder_ slot is used with any other field than `ContentTypeId` it will break. It should support other boolean properties as well (or rename the slot `ContentTypeId` instead and mention in the doc this value is only used to generate preview URLsm like `ListId`,`ItemId`, etc.)

search-parts/src/webparts/searchResults/components/SearchResultsContainer.tsx
FranckyC: In the end, if the URL is still not correct for end users, they can simply update the template logic themselves. That is the whole point of these Web Parts ;). We should only cover well known scenarios the best we can.

wobba: Path is not a good one for files items imo. You typically need some logic to pick the better URL.

search-parts/src/webparts/searchResults/components/SearchResultsContainer.tsx
wobba: @FranckyC Good point. Change this and the one above to always get the slot from the dataSource default slots.

PR #1317, opened 2021-09-19
Pick language from page translation over page context if available

wobba:


search-parts/src/dataSources/SharePointSearchDataSource.ts
```diff
@@ -190,6 +191,15 @@ export class SharePointSearchDataSource extends BaseDataSource<ISharePointSearch
         }
     }
 
+    private getTranslatedCultureFromUrl(): string {
```
FranckyC, line 194: Any example strings that match this exp? From what I see it, any page within a folder will match this resulting to an incorrect language (ex: "/SitePages/test/test.aspx"). The regex should be more precise to match a culture format.

wobba: Good comment, but for modern you don't have pages in Folders... unless you hack it in. So still a concern?

wobba: You can make folders, but not make modern pages inside them. Committed fallback if culture resolves to 0.

FranckyC: image

It seems you can :D

wobba: They keep changing things :D I don't have "create folder" as an option in my SitePage lib as default. Either way, the fallback should work just fine. It's not intended to create folders 😉

FranckyC: > 
> 
> They keep changing things :D I don't have "create folder" as an option in my SitePage lib as default. Either way, the fallback should work just fine. It's not intended to create folders 😉

Folders are used for security reasons on navigation links, that is the only reason imo. Anyway, just consider this scenario as well in the regex

wobba: Changed the logic to take the last folder name as culture candidate. Tested and verified with subfolders and translations 👍

FranckyC (approved):

PR #1331, opened 2021-09-22
Clarify how to add managed properties to the selected properties in search results.

juliemturner:


wobba:
Thank you and about time we added this :) I've also changed the code in the dev branch to allow pasting a comma separated list as well, not just adding individual ones. Could you include that as well?

wobba:
Fixed

wobba:
Manually merged with #1345

PR #1342, opened 2021-09-25
Adds support for dynamic sort in details list 

ellinge:
Added the possibility for dynamic / query sort in details list.

Changed the checkbox to a dropdown for sortable:
setupdynamic

And the sorting in action (title has been replaced by contenttypeid).

- Sharepoint search with initial sort on created
- Ms search search with initial sort on created
- Sharepoint search with no intial search

intialsorting

I'm adding it as a draft just to start the discussion:

What's missing:
- ~~CollectionDropdownField https://github.com/pnp/sp-dev-fx-property-controls/blob/7fb3a42fcf61e69c49b7ba589dfa941b7ada62cd/src/propertyFields/collectionData/collectionDropdownField/CollectionDropdownField.tsx#L77 does not show false as possible value. I wanted it to be backward compatible with existing settings. Should make a pr to allow false as a valid key.~~ https://github.com/pnp/sp-dev-fx-property-controls/pull/397 
- ~~validateSortableProperty seems to return true for Title but I get a 500 error when trying to sort it dynamically, it it a bug? The same happens if I try to sort it with the existing settings.~~ I increased the number of items in the validation and it triggered an error for Title and FileType now.
- From what's available in getPropertyPaneFieldsConfiguration I can't find a way to know what data source is used. I could check for Rank vs rank I guess but it would be better to do it another way. Include some more context with validateSortable on datasource instead to have each implemention done there?
- The sortable field for ms search gets prefixed mostly with "resource.". I have to manually change the property to createdDateTime for it to work as a sortable. Should one just strip the available fields of this value?

Improvements:
- ~~Should the sorting be aware if default settings for sort has been set? Should it be reflected in initial render?~~
- Could add a filter connection and a sort webpart for other scenarios
- Different icon depending of type of sorting?
image


ellinge:
I've made it aware of current sorting
- Sharepoint search with initial sort on created
- Ms search search with initial sort on created
-  Sharepoint search with no intial search
intialsorting

wobba:
In the edit experience, should the header be on top, not per dropdown? Looks a bit off with alignment.

ellinge:
It's as requested in https://github.com/microsoft-search/pnp-modern-search/pull/1282.

Unfortunentely there's out of the box event to toggle column visiblity so the label was set to '' to "hide" it when there's no column with handlebars + sorting enabled.

I've made a PR https://github.com/pnp/sp-dev-fx-property-controls/pull/394 to allow column visibility to be dynamic instead using empty labels.

wobba:
v1.0.5 is now updated in npm

wobba:
We'll process this for the next version.

wobba:
Want to update that this one is still queued up to me merged.

FranckyC:
@ellinge can you resolve the merge conflicts with the latest develop branch please? Sorry for delay processing.

ellinge:
Updated now, I also updated @pnp/spfx-property-controls which has some simplified methods now for column visibility and custom validation https://github.com/microsoft-search/pnp-modern-search/pull/1342/commits/203aacf1e88309f616936e6c3e6e1a00d43cf234

search-parts/src/layouts/results/detailsList/DetailListLayout.ts
```diff
@@ -2,12 +2,15 @@ import * as React from "react";
...
 import * as strings from 'CommonStrings';
 import * as propertyControlStrings from 'PropertyControlStrings';
-import { Checkbox, ICheckboxProps, IComboBoxOption } from 'office-ui-fabric-react';
+import { IComboBoxOption } from 'office-ui-fabric-react';
 import { IDetailsListColumnConfiguration } from '../../../components/DetailsListComponent';
 import { IPropertyPaneField, PropertyPaneToggle, PropertyPaneDropdown, PropertyPaneHorizontalRule, PropertyPaneButton, PropertyPaneButtonType } from '@microsoft/sp-property-pane';
 import { TemplateValueFieldEditor, ITemplateValueFieldEditorProps } from '../../../controls/TemplateValueFieldEditor/TemplateValueFieldEditor';
 import { AsyncCombo } from "../../../controls/PropertyPaneAsyncCombo/components/AsyncCombo";
 import { IAsyncComboProps } from "../../../controls/PropertyPaneAsyncCombo/components/IAsyncComboProps";
+import { SharePointSearchService } from "../../../services/searchService/SharePointSearchService";
```
wobba, line 11: Taking a dependency on search service in this component is the wrong approach. Take a look at how e.g. pagination is implemented. As the data source is agnostic on the picked source (Graph or SharePoint).

ellinge: https://github.com/microsoft-search/pnp-modern-search/blob/203aacf1e88309f616936e6c3e6e1a00d43cf234/search-parts/src/layouts/results/detailsList/DetailListLayout.ts#L332

It has with validation of sortable to do. It’s not provided for graph, it was not at least not back when this was done. There’s also a question on the pr about the returned column names for graph. I had to forcefully set another name to get sorting to work if I remember correctly.

ellinge: The same is done here:
https://github.com/microsoft-search/pnp-modern-search/blob/ddaa763d1e0b5f0ed8e5d0ebd9c7c7d8d8fdba7e/search-parts/src/dataSources/MicrosoftSearchDataSource.ts#L19

ellinge: Same is done for graph here:
https://github.com/microsoft-search/pnp-modern-search/blob/ddaa763d1e0b5f0ed8e5d0ebd9c7c7d8d8fdba7e/search-parts/src/dataSources/MicrosoftSearchDataSource.ts#L307

Currently that’s not exposed in a agnostic way. Which I tried when following the implementation for pagination.

FranckyC: Actually, the sort properties and validation should be at data source level and passed to different layouts using data attributes in the Handlebars templates. Layouts shouldn't have direct references to services. The only thing that they should do is raise the sorting event back to main Web Part so the data source can retreive new results and re-render the layout.

It means, dynamic sorting feature should be available for all layouts and not only the DetailsList so the design has to be more generic.

Also, if we add such feature, static sorting is not relevant anymore so no need to add an additional option in column configuration.

ellinge: The reason why it’s part of the DetailsList layout it’s because that’s the only layout currently that supports custom sorting triggered by the end user in the first place. 
Also, the static sorting is not obsolete because of this. Sorting is more of an exception than a rule when it comes to managed properties from what I’ve seen. Title for instance is not sortable out of the box. You will have to bind the values to a refinable string instead. This is the reason the validation is done in the first place.
The data source properties are populated here https://github.com/microsoft-search/pnp-modern-search/blob/ddaa763d1e0b5f0ed8e5d0ebd9c7c7d8d8fdba7e/search-parts/src/webparts/searchResults/SearchResultsWebPart.ts#L576
And the “template” properties are populated indirectly from here https://github.com/microsoft-search/pnp-modern-search/blob/ddaa763d1e0b5f0ed8e5d0ebd9c7c7d8d8fdba7e/search-parts/src/webparts/searchResults/SearchResultsWebPart.ts#L602 (calls getLayoutTemplateOptions)

So not really a clear approach on how one would bind the columns displayed in the details list with a setting on the layout level without a possibility to perhaps override the properties for the template from the data source (a.k.a change the fields for layoutProperties.detailsListColumns adding dynamic as an option + validation for valueSorting). It would be better to have the validation callable in an agnostic way so that Graph for instance could check that the type of query matches a sortable one etc.

wobba: @FranckyC if we remove the validation, do you still think it's a custom web component or can we merge it? And the dropdown should be a text box probably as well - or it could be drop down for the SP source if possible.

FranckyC: I'm working on a sort control more generic using a component to be integrated to all layouts. I may use parts of code you did @ellinge for the details list however. I'll put this PR on hold then.

PR #1357, opened 2021-10-03
Feature/export to csv

### LINKED ISSUE #503: Feature to have export functionality for search results
**Is your feature request related to a problem? Please describe.**
We are using pnp modern search web part for document library search, we would like to request if there is a provision to export the search results for a specific keyword in the existing webpart or is there any future plan to incorporate the export functionality

**Describe the solution you'd like**
I have found a link while searching for this requirement:
https://social.technet.microsoft.com/Forums/en-US/820e088d-3185-45a8-9361-045cc449bab8/export-search-results-to-excel?forum=SP2019

image

We would like to have a feature as shown in this picture above , not sure from where they have achieved it..as per their comment its an OOB solution, not sure if it's really is.

**Describe alternatives you've considered**

**Additional context**

ellinge:
Adds a draft of exporting to CSV. Fixes #503 

exportascsvtest

The configuration:
image
image

What's left:
- Not sure it works with other cultures but export with Swedish åäö seems to work fine
(Added BOM and changed separator to ";" otherwise I got the UTF-8 encoded file to be read as ANSI in excel)
- ~~Add a dialog before exporting selecting current page or all (visible if total results > items.count). Probably should have some limit though. Use docid-approach as paging while fetching for large result sets?~~
- ~~Dialog could have file name with a prefilled one based on title of search result web part and current date or something similar?~~ Skipping this, changed to info dialog instead
- Does not care about sorting for now. Could fetch .items if users selects current page but that would mean no static sorting would be applied if not done again probably.
- ~~Style button. Flex it with the other options and put it to the right?~~
- ~~Localize texts for button label and texts explaining limitations of sorting and max result set~~
- ~~Browser check for support? Do we need this? Just disable it if link.download not available?~~ Added text if a.download is not available.
- ~~Default values set when enabling export?~~
- Code refactoring
- What to present if search api returns 500 or a column is not handlebared correctly? Set column value to ERROR? Show red bar with more info in information dialog? Show MessageBar with error?


ellinge:
Changed dialog to settings intead, had problems triggering download correctly when triggering it from dialog for some reason.
Probably remove the selection of Current page / all?

wobba:
Does it export the current result set/page only, or does it kick off a query with paging?

ellinge:
It can do both.

See:

ellinge:
I’ve limited the result set to a maximum of 10k items.

wobba:
Can you retrieve indexdocid as part of your results and use this? At least for SPO content. Connector content might be different. Or we use a different approach with the SPO and Graph provider until we know how best to make it work. 

I posted a question at https://docs.microsoft.com/en-us/answers/questions/590157/what39s-the-best-way-to-paginate-large-data-set39s.html to see if there is any guidance.

ellinge:
Did a try with the docid-approach now, see latest commit.
### Compared results with same sorting applied for fields in excel:
image

### Single, DocId-approach
Processed '1', items total fetched 500
Processed '2', items total fetched 1000
Processed '3', items total fetched 1500
Processed '4', items total fetched 2000
Processed '5', items total fetched 2500
Processed '6', items total fetched 3000
Processed '7', items total fetched 3500
Processed '8', items total fetched 4000
Processed '9', items total fetched 4500
Processed '10', items total fetched 5000
Processed '11', items total fetched 5500
Processed '12', items total fetched 6000
Processed '13', items total fetched 6500
Processed '14', items total fetched 7000
Processed '15', items total fetched 7500
Processed '16', items total fetched 8000
Processed '17', items total fetched 8500
Processed '18', items total fetched 8922
fetching: **31491.22607421875** ms
mapvalues: 1942.12109375 ms
Processed 'TestWithFilename 2021-10-14.csv', items total exported 8922, has error column value: false, empty rows: 0
ExportComponent.tsx:173 exporttocsv: 34.800048828125 ms

### Parallel, no DocId-approach
Processed '1, 2, 3, 4', items total fetched 2000
Processed '5, 6, 7, 8', items total fetched 4000
Processed '9, 10, 11, 12', items total fetched 6000
Processed '13, 14, 15, 16', items total fetched 8000
Processed '17, 18, 19, 20', items total fetched 8922
fetching: **23146.856201171875** ms
mapvalues: 1911 ms
ExportComponent.tsx:173 Processed 'TestWithFilename 2021-10-14.csv', items total exported 8922, has error column value: false, empty rows: 0
ExportComponent.tsx:174 exporttocsv: 36.455078125 ms

rameshbaburanga:
Hi All,
I am unable to get export to csv option on PnP search results. we deployed the PnP Modern Search - Search Web Parts - v4

rameshbaburanga:
Hi All,
Kindly suggest me here.
I am unable to get export to csv option on PnP search results. we deployed the PnP Modern Search - Search Web Parts - v4

ellinge:
It has not been merged/released. You would have to build it yourself from this branch for it to appear. But I have no idea if or when in time this would be released. If ever, so then you would be stuck with that build. So I wouldn’t really recommend it. You could theoretically rename the package and webparts before build to be able to have they both side by side, but I guess that would be a bit confusing as well.

rameshbaburanga:
Hi All, 
Kindly help me on below issue.
Where can we see the Search - Search Web Parts - v3 log information.

search-parts/src/components/DetailsListComponent.tsx
```diff
@@ -514,35 +484,6 @@ export class DetailsListComponent extends React.Component<IDetailsListComponentP
     return groups;
   }
 
-  private _processHandleBarsExprValue(columnValue: string, item: any): string {
```
wobba, line 517: Can you explain the refactoring here... I see the code is a lot simpler. Just to understand why the old version was too verbose.

wobba:
Am I reading the code correct that there is one extra web component - and the other changes are refactorings which are not related to the export web component?

search-parts/src/components/DetailsListComponent.tsx
ellinge: It was done in several places before. It’s been moved to (for reuse): https://github.com/microsoft-search/pnp-modern-search/pull/1357/files/0bf0b54502c87b10dd641a54dc0865c0e596a9f4#diff-0f3fa7a21909dd1252a21cc87075c10699d2bb4a7550d4fcedc7cfc7fadfbd92R15

it was needed for the export component and didn’t want to to a new copy paste.

ellinge: Also I made it possible to separate the template compilation (once per column instead of column+row) from the value extraction to speed up the process when handling a large quanity of data.

ellinge: The only thing to have a though about is that i one instance you specified which properties which was sent to the template and other times you use the spread operator on the whole templateContext:
image

Is there's a performance boost of deciding on what to do here?

ellinge:
Finnish seems to be on its way also 😨 

The refactorings are done:
•	to be able to make use of the data source outside the search results webpart. Since it's a separate query with other possible fields.
•	for reusing the Handlebars template logic. And being able to fetch the compiled template once per column
•	to store common strings as instead on each template type.
I think that covered it.
I will update the info text displayed to the user as well, cover Norwegian and add the documentation when I find time.

As you can see, I got the same results faster using parallel queries compared to docid sort even if each query itself seems quicker. I have two support issues about the search (with docs appearing and disappearing without any change and wrong sort order for åäö) so I don’t expect the search result to be that reliable either way… So, a disclaimer should suffice.

ellinge:
Have updated the language files.

Have updated the documentation for details list. Not sure if the web component should be included in https://github.com/microsoft-search/pnp-modern-search/blob/develop/docs/extensibility/web_components_list.md ?

Is that really a reusable component? We probably could add support to other templates later on so that <pnp-export is used in more placed but currently it's quite tided to the DetailsList (since that's the one displaying the most tabular data). <pnp-filter* etc are not listed here.

wobba:
If it's a web component, then it should be able to use for any layout - manually added.. right? Grabbing the current query, and paging it.

ellinge:
Yes but the settings is only enabled for detailslist for now. I’ve built it to be generic though. There’s settings on which columns should be rendered. Since we shouldn’t support html and such for this. But I guess one could fallback to the display columns if no export is defined.

wobba:
I see your point. Looking at the input props you cannot just drop it in any random template and it would work. Question is, can it be generalized to make this work? E.g. implicitly fetch the service scope and context without passing it as an example?

FranckyC:
Hi @ellinge, with @wobba we are wondering if it was not better for component to be a standalone component available as a custom extension and not tied to the core logic. Could you try to "externalize" it there https://github.com/microsoft-search/pnp-modern-search-extensibility-samples following the same pattern?

This shouldn't be tied to the DetailsList layout specifically in my opinion (or any other layout). Normally, you should have all the required information from the Handlebars context to replicate the same exact query for export as the configured data source. Also, why not create an UI setting to select columns to export manually instead and make generic for all layouts? 

As you may understand, we are trying to limit code modifications in the main code base to improve long term maintenance. By doing this as an extension, you could maintain and improve it separately with no strong dependency with main solution.

Thanks!

ellinge:
Even if one should externalize the export component, there needs to be changes made to this library for reuse of logic regarding data source and handlebars. Eg. manifestVersion that is used for the graph api is inserted from the webpart props and not accessible in the template (and therefore I would not be able to recreate a dataSource that's working with the graph api without this information).

I've also extracted the dataSource-initalization to a separate helper to be able to be able do a paginated query against an own dataSource and change what props are returned based on the users selected export properties.

When it comes to the handlebars logic that does not work well when performing it on many rows, so I've also extracted this to a separate handler for reuse / DRY.

Since the managed columns for the table is also able to handle html, probably nothing you want I excel, and you probably would want to include more information in the exported result? So either way there’s a separate setting for the export columns. These are also separated into another class to be able to resuse it for other templates. This could of course also be "bumped" up to be included everywhere of that's what you mean by "generic for all layouts". It would make no sense to have these settings without the export component being part of the library as well though.

From what I can see there's an interest from many people to have this kind of feature in the component. And we have a build running for a customer. It would be nice to avoid having to do a custom build for them.

I'm not sure how you propose to expose the dataSource but let me know and I will see what's possible. Otherwise I will have to just fork this and let it be custom I guess. Is the extensibility library also being able to add new settings or how would that be solved? I only thought it meant registering a new <pnp-component for use in custom templates.

FranckyC:
Actually you don't need to recreate the data source. You need to do the exact same call as the data source that is not the same thing. For that, you already have the SharePointSearchService available in web components through the default service scope and all data source properties from the Handlebars context to be able to build the search query. The MicrosoftSearchService could be easily exposed to web components as well.

For the output, actually exporting search results is different from exporting the current results "layout" including all field transformations in my opinion. To make it reusable for all layouts, the export feature should only export raw data and not UI processed data, so the Handlebars processing methods are not needed here because they are too specific depending the chosen layout.

The columns you need to export are basically the ones in the "Selected properties" option (we could also expose the list of current available results fields in a separate context property if needed). Also, you could add a setting in your component to choose what fields to include in the export based on that list including the number of pages you want to export (or any other setting you want to customize the export query).

As it is implemented, the code is too invasive for a single feature and can't be merged as is.

So to summarize:

- Don't recreate the data source, just use the search services with the same settings in your component. If shared helpers are needed, I'm fine with this.
- Export the raw data, not the UI processed results as it would be too complicated.

Just that is a good first version.

ellinge:
First off, have you tried the feature out?

Secondly, in what way is it too invasive? Is it:
-	the refactoring done? The reason for extracting the handlebars template extraction is because currently it is repeated in 3 different places and for me to able to precompile the template once for reuse on each row/column.
image
Regarding the new instance of the/exposing the initialization of dataSource it’s because the fields set for retrieval is done on creation so fetching 10000 items would mean fetching a lot of junk data probably. Not sure how much of a difference it would do to the performance though. I’ve also tried to reuse/share some common labels across the different templates in the language resources.
Other than that most of it is isolated to the Export*-files.
-	adding new properties to the component? One could instead export everything checked in Selected properties, or the Manage columns or all fields available in the response. The reason I choose not to export Select properties of fields properties raw is because the end user would not care or know what to do with half of it. There would be cryptic names and no logic in sorting between columns. It’s the same reason I don’t think the end user should see a list of these fields for selection either. And as I’ve said before, the Manage columns supports things like html which and you probably would want more info exported than what’s visible. 
-	a new feature to maintain? For from what I understand this seems to be quite a common wish. I investigated the extensibility at first but did not see it as a viable option, currently, for bringing a user-friendly experience for this. It would require more options as extending properties to store settings and exposing a few more things in my opinion

You say it would be too complicated, for whom, maintainers? I argue it would be more complicated for the end user if you would just export the data raw, or having the user select properties from a list of managed properties or fields. Should they do this every time they do an export, or should this selection be stored somewhere, localStorage? And having it all exported as raw means a lot of data washing would be having to be done in Excel. Sort columns, removing irrelevant data, rename column headers, reformat date etc.

image

image


From my perspective, if not keeping it forked, the common ground to be found would be to only manually being able to include this in a template (pnp-export, preferably as an extension sample?) and provide it with a attribute harcoded with headers/fields to export always (no selection for the end user)? One would propably would have to always parse dates in the export as default as well. The dataSource could be reused from the one initialized in the search results web part. I would still like to see it being able to change what properties to return with the data context as with the dynamic filtering/sorting though to limit what needs to be fetched, both having sp servers in mind and the processing time neeeded. It think this would be more complicated for the user setting up the export though in my opinion. But I guess this is not something that's setup that often at the same time.

wobba:
I've tried it and it solves the use-case just fine.

However, instead of adding all edge-case functionality into the core part we want to use the extensibility model as much as possible. This means those needing the functionality can install it and add the web component to a custom template. We can add a separate repo for extra web components (haven't so far because no one has contributed any) and I think this one is a perfect fit.

Extensibility should be thought of as a plug-in model where you can install extras when needed - as export to CSV is not that common a scenario.

If we turn this into a custom web component following https://microsoft-search.github.io/pnp-modern-search/extensibility/custom_web_component/ you should be able to get hold of the search and handlebar service scopes (or we expose the ones not there) so you can consume them in the web component.

I'd be happy to give it a go, but will be some time before I get to it, so I'd like to keep this PR for reference. Changing it to draft for now.

FranckyC:
> First off, have you tried the feature out?

Yes I tried it and reviewed the code. Conceptually, web components don't know and shouldn't know anything about the data sources (as it is an internal concept), they can only use exposed generic services. What you are doing in the `ExportComponent` component with `DataSourceHelper.getDataSourceInstance` is basically an anti-pattern because you basically create a new instance of the data source on the first export every time (at page load, and every single Handlebars re-render). This is not the way service scopes are supposed to work. In addition, you forgot to pass the instance id in the component to get the Web Part specific scope (`data-instance-id="{{@root.instanceId}}"`) so you are creating a new child service scope every time. Except the `getData` method you don't need any of other methods of the data source. I stil maintain you can just build the same search query and use the `SharePointSearchService` to achieve the same goal with less dependencies. Data source is an internal concept not meant to be manipulated by web components.

As stated in the documentation for extensibility:

>  all builtin data sources, layouts, web components or suggestions providers are built using the same exact interfaces and methods that are publicly available in the @pnp/modern-search-extensibility SPFx library project.

Meaning you can't bypass this design just for your feature and use a `DataSourceHelper` or any other helper just for this specific use case.`DataSourceHelper` and others are not available in other web componentns made as extensions.

> There would be cryptic names and no logic in sorting between columns?

You can always manage this as configuration otpions for your component (not in the property pane) as well as format options (ex: for dates).

image

> For from what I understand this seems to be quite a common wish. 

Not really no.

> Should they do this every time they do an export, or should this selection be stored somewhere, localStorage?

Basically yes it is an option until we figure out how to allow persistence for web components ex: custom layouts)

> You say it would be too complicated, for whom, maintainers?

Yes as we are responsible to ensure everything works correctly. We had a pretty painful experience with v3 ;)

ellinge:
I really wished you would have responded from the start with this information. It makes perfect sense now that it's bad practice the way I implemented disregarding if it was part of the main lib or as an extension. Either way, my proposal to getting this forward would be the following:

- You create a new place where implemented ready-to-use extensions, like this one, can be hosted and maintained (as suggested by @wobba) I think it would be good if releases were provided and not just sample code which you need to build yourself.

- The extensibility library is extended with the possibility to inject more properties to the web part (aka export and manage columns becomes available “everywhere”). These settings could be stored in the webpart json under extensibility: { ‘package-name’: { settings… } and exposed in same way under dataSourceProperties for use in templates etc. These settings could be shown in a separate panel in the config.

- The export manage columns could be extended with a “export as default”-checkbox and you could create settings for several columns you think the end user could want to export without them having knowledge of managed property names or how to parse/format the values. One could then add a “selection of columns to export” with the “export as default” checked by default to the end user in the export component.

- Revert the expose of datasource-initialization from search results and fetch the instance in the ExportComponent. One could probably extend the dataContext passed with selectProperties to getData to be able to fetch other set of columns. This is a perhaps. Not sure yet how much of an impact a lot of columns not needed for the export would cause.

How does that sound?

FranckyC:
I would suggest to process step by step to get a minimal working version first and validate the approach:

1. Move your component to this repository https://github.com/microsoft-search/pnp-modern-search-extensibility-samples 
2. Use existing `@pnp/modern-search-extensiblity` and web component `serviceScope` capabilities to do a **raw export** using existing Handlebars context properties (`properties.dataSourceProperties`) and `SharePointSearchService`. If you need more context properties along the process, update the core code (only if absolutely necessary). Basically, the default columns to export are the default search selected properties.

image

3. Make it work for **all** layouts when the extension is deployed. To do that just customize existing layout templates and add your export component.

When all of this functional, we will figure out how to make it more integrated with the base solution.

FranckyC:
@ellinge I've added custom layouts support for extensibility. Please take a look at https://microsoft-search.github.io/pnp-modern-search/extensibility/custom_layout/. This way you could create a layout dedicated to export + save your settings in the Web Part property bag.

PR #1432, opened 2021-10-22
#1367 - Added filter deep links feature

FranckyC:
Added filter deep links feature. Now the selected filters are set in the URL dynamically as query string parameter so you can use them later to pre-filter results. It means the refiners Web Part is now a kind of builder to set up your filters as deep link.

@wobba can I have your feedback/tests on this? It is a pretty simple implementation but should work for all scenarios.

### DIFF
--- a/docs/usage/search-filters/index.md
+++ b/docs/usage/search-filters/index.md
@@ -69,6 +69,8 @@ These taxonomy values can then be used in the Filters Web Part using a `Refinabl
 
 The Search Filter Web Part supports deep linking, meaning you can preselect filters from the URL at page load. When filter values are selected, a query string parameter `f` is append to the current URL containing the current filter values data.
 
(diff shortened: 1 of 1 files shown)


raymond-tishenko:
Thank you for the work on this item! We'll get to testing this using the latest build and let you know how it goes.

PapaGeil:
Just stumpled upon this feature and it is very useful. Thank you for having it implemented a while back. 

I am struggling with one thing, though. I want to dynamically create URLs with certain filters in my Search Results web part (I know, the documentation tells me not to do this, but the syntax of the deep links is quite straight forward). I am wondering whether there's a way of encoding a given filter string to Base64, so that it's compatible with the URL syntax? I checked on the documentation, but couldn't find anything. The company I'm working for doesn't allow me to write my own javascripts, so I'm hoping for a helper of some kind. 

Thanks!

wobba:
@PapaGeil if you have a question ask it in the Discussions area.

PapaGeil:
@wobba sorry, will do

PR #1455, opened 2021-10-28
Upgrade spfx 113

wobba:
Keep as draft until bundle size has been investigated


FranckyC:
@wobba maybe we should wait for 1.14 instead?

wobba:
We can..let's leave it in draft for reference when we get there. No rush at all 🙂

search-parts/package.json
```diff
@@ -71,7 +71,7 @@
         "colors": "1.4.0",
         "fancy-log": "1.3.3",
         "gulp": "4.0.2",
-        "spfx-fast-serve-helpers": "^1.12.27",
+        "spfx-fast-serve-helpers": "^1.13.0",
```
s-KaiNet, line 74: Sorry for my 50 cents, but here we need `~`, i.e. `"spfx-fast-serve-helpers": "~1.13.0",`, so that only the patch version is updated on fresh npm install.

wobba:
@s-KaiNet the main reason why I haven't upgraded is that the bundle size increased as webpack chunks it differently. I haven't had time to properly investigate so we might wait for 1.14 until we upgrade. I'm leaving this in draft until I figure it out.

Issue #1429, opened 2021-10-21
Search Box: "undefined" when empty query parameter


till-llit:
**Version used**
4.4.0

**Describe the bug**
Search box will display "undefined" when query parameter is null or empty

**To Reproduce**
1. Configure PnP Search Box webpart to use a page environment -> query string -> query parameter
2. Load the page without the query parameter
3. Search box is now prefilled with the text "undefined" 

**Expected behavior**
Search box will display nothing or the placeholder text

**Screenshots**
![image](https://user-images.githubusercontent.com/91128746/138228379-853cd937-ece7-4ee5-96c5-fb113e2711a6.png)

![image](https://user-images.githubusercontent.com/91128746/138228437-240c16c7-4931-4a9f-a43c-7438463e2ae9.png)

**Desktop (please complete the following information):**
 - Browser Edge Chromium

**Additional context**
not reproducable with 4.3.0

PR #1466, opened 2021-11-02
Amend fix for #1429

phoob:
This fixes situations where search query could be string('undefined') if the dynamic source for the input query is set to page context -> search.


wobba:
Thank you for adding this as I forgot 👍

PR #1468, opened 2021-11-02
Add getTagName helper

phoob:
Unless tags are mapped to a refinable string, they are rendered as `L0|#00033d7f8-e002-439b-a6fa-2d1a5a92e1bf|TestTag`. Would this be an ok solution to get the tag name? Instead of writing `{{last (split name "|")}}` every time?


wobba:
It would, if you also add it to the documentation as part of the PR :)

phoob:
Ok @wobba , was that it?

wobba:
Thank you :)

Issue #1467, opened 2021-11-02
Query Template for Microsoft Search


edarroudi:
**Is your feature request related to a problem? Please describe.**
I need to be able to add a prefilter to the Search Results webpart. This is possible for the SharePoint search with the "Query template" property but not for the Microsoft Search data source. 

**Describe the solution you'd like**
It would be nice to add the query template property for Microsoft Search data source as well.
![image](https://user-images.githubusercontent.com/935493/139811029-97037d79-4406-455d-994a-89ef9ef9e0cb.png)
I did develop this myself and it works fine.

**Describe alternatives you've considered**
Using filters wouldn't work, i thought of a custom searchbox component that sends the modified query to the search results webpart but that doesn't make any sense.

**Additional context**
I have a task to modify & add some additional features to the modern search webparts. I consider some of the modifications as "too company specific" but some of the changes could be of use for the global community. How should i approach this, a) upload a list of possible features and you vote which ones can be implemented by me (maybe with some modifications) or b) just create pull requests and you have a look at those modifications?

FranckyC:
Hi @edarroudi , unfortunately, there is no query template option for the Microsoft Search API for now. We can't do much about that :/. For the features you want to add, please describe them in 'Ideas' section in the discussions tab https://github.com/microsoft-search/pnp-modern-search/discussions/categories/ideas.

edarroudi:
> Hi @edarroudi , unfortunately, there is no query template option for the Microsoft Search API for now. We can't do much about that :/. For the features you want to add, please describe them in 'Ideas' section in the discussions tab https://github.com/microsoft-search/pnp-modern-search/discussions/categories/ideas.

hey @FranckyC , 
Thank you for your reply.
I actually don’t specifically need a querytemplate but just a way to enhance /modify the user query as shown in my screenshot. I modified the code and  just pass it in the query string and that works fine. Would that be a bad feature? I don’t see  an issue here.

wobba:
@edarroudi I think that makes sense. Have the option to specify additional KQL in the web part which is appended to the user query.

edarroudi:
Hey @wobba , thanks. 
As i already coded that (have to refactor it a bit), is it ok for me to create a pull request? I'm new to this kind of community driven coding so sorry for any stupid questions.
Would you need some specific naming for the label? I used the label Query Template but maybe the naming will cause some misunderstanding. How about "KQL Query modifier"?  

[...]

PR #1484, opened 2021-11-04
Microsoft Search: KQL query text Extension

edarroudi:
Added a property 'Query modifier in Microsoft Search DataSource 
image

Added labels for the different languages through google translate.

It just attaches whatever is defined in the property to the queryinput text.
e.g. if template is set to 'AND IsDocument:true',  a search with the input string 'security' will be transformed to 'security AND IsDocument:true'

The template accepts tokens as done in SharePointSearch:QuerySearchTemplate.

Was discussed in issue : https://github.com/microsoft-search/pnp-modern-search/issues/1467


wobba:
I think you should omit `AND` from the sample, as AND is implicit anyways in KQL.

Curious if we should add `{searchTerms}` as a concept and then replace it with the user query if present - to make it consistent from a web part point of view.

Any comments @FranckyC?

edarroudi:
> I think you should omit `AND` from the sample, as AND is implicit anyways in KQL.
> 
> Curious if we should add `{searchTerms}` as a concept and then replace it with the user query if present - to make it consistent from a web part point of view.
> 
> Any comments @FranckyC?

I did have it like this in my earlier version, but as {searchTerms} isn't part of the tokenService and i had just a regex replace for it in code it looked out of place. 

I'll remove the ADD part, forgot the implicity, thanks.

edarroudi:
found a setting error then in context if we go the way right now without {searchTerms}. allowEmptyValue has to be true.
I will fix this and the rest when you guys decicde what your want :)

edarroudi:
Hey @wobba, 
did you decide what you want: either including {searchTerms} in the modifier template or as it is right now, concat it to the inputModifier. Please tell me if i can do something else to get this approved ;-)

pettersh:
In my opinion I think the property label should be **Query template** instead of **Query modifier** for consistency. 

I understand that this feature is a query modification, but that's what SharePoint query templates does under the hood as well; modifies the user's query before it's sent to the search index.

edarroudi:
@pettersh If i rename it to Query template then i have to go the way @wobba described, include {searchTerms} in the Query template - for consitency reasons :) I think that would be better anyways for overall consistency.

wobba:
@edarroudi did you implement my ask? I think end-user familiarity is more important than the actual technical implementation. It also allows easier migration from SP API to Graph API.

edarroudi:
@wobba  sorry - didn'T see your thumbs up. I'll do it till monday.

search-parts/src/dataSources/MicrosoftSearchDataSource.ts
```diff
@@ -468,12 +502,19 @@ export class MicrosoftSearchDataSource extends BaseDataSource<IMicrosoftSearchDa
...
         // Query text
         if (dataContext.inputQueryText) {
             queryText = await this._tokenService.resolveTokens(dataContext.inputQueryText);
         }
 
+        // Query modification
+        const queryModifierTemplate = await this._tokenService.resolveTokens(this.properties.queryModifierTemplate);
+        if(queryModifierTemplate)
+        {
+            queryText = `${queryText} ${queryModifierTemplate.trimLeft()}`;
```
FranckyC, line 515: Wrong statement, you should have the same behavior as the SharePoint search. The query text should be included on demand using either `{searchTerms}` and `{inputQueryText}` tokens. It means you should add `{searchTerms}` as a builtin token and resolve it the same way as `{inputQueryText}`.

search-parts/src/dataSources/MicrosoftSearchDataSource.ts
```diff
@@ -59,7 +60,14 @@ export interface IMicrosoftSearchDataSourceProperties {
     contentSourceConnectionIds: string[];
 
     /**
-     * Flag indicating if the Microsoft Search beta endpoint should be used
+     * The search modfier template. Extends the query with additional KQL
+     * e.g. if tempalte is set to 'IsDocument:true' a search with the input string 'security'
+     * will be transformed to 'security IsDocument:true'
+     */
+     
+    queryModifierTemplate: string;
```
FranckyC, line 68: Rename to `queryTemplate` to align with SharePoint Search

search-parts/src/dataSources/MicrosoftSearchDataSource.ts
```diff
@@ -451,6 +483,8 @@ export class MicrosoftSearchDataSource extends BaseDataSource<IMicrosoftSearchDa
         this.properties.fields = this.properties.fields !== undefined ? this.properties.fields : CommonFields;
         this.properties.sortProperties = this.properties.sortProperties !== undefined ? this.properties.sortProperties : [];
         this.properties.contentSourceConnectionIds = this.properties.contentSourceConnectionIds !== undefined ? this.properties.contentSourceConnectionIds : [];
+        
+        this.properties.queryModifierTemplate = this.properties.queryModifierTemplate ?? "";
```
FranckyC, line 487: Use the same initialization as SharePoint Search `this.properties.queryTemplate = this.properties.queryTemplate ? this.properties.queryTemplate : "{searchTerms}";`

FranckyC (changes requested):
Please update according to comments before merging. Also check with the current `develop` branch and add missing translations if any. Thanks!

FranckyC:
Hi @edarroudi, I've made the changes in the PR #1621 because I needed them before the December release. I'll close this one.

Issue #1246, opened 2021-08-30
Dynamic Constants for Microsoft EndPoints


PwnGinger:
**Is your feature request related to a problem? Please describe.**
Our environment is located within GCC-High and not the Commercial space. Due to this, a majority of the endpoints leveraged for API calls, example GRAPH, aren't ".com" but instead ".us". Within the source code, there are hardcoded Microsoft URLs that lock in the Commercial endpoint instead of dynamically determining the space and leveraging .us.

Example from the code: const MICROSOFT_SEARCH_URL = "https://graph.microsoft.com/beta/search/query";

**Describe the solution you'd like**
Ideally, any constants leveraged for URLs should be dynamically generated. At least, a switch/conditional should be leveraged to allow non commercial endpoints to be used for GCC, GCC-High, or DOD tenants.

**Describe alternatives you've considered**
Manually adjusting the code to account for our GCC-High endpoints. While this should work, we would need to do this for any future package updates. Since this issue would impact any other customer within a non-commercial space, I believe this should be handled on y'alls end.

**Additional context**
None

wobba:
I'll add this as an enhancement to be included. I don't believe Graph Search is available yet in GCC-High, but we should still add the support to allow changing the environment when needed.

Which mean you have to use v4 and the SharePoint provider for now.

PwnGinger:
Yeah, we have no issues with the SharePoint results. So, we can just leave the results webpart set to "SharePoint". Microsoft notes that the Graph API isn't available in GCC-High yet, however,  we do have bits and pieces that work today.

I appreciate the quick response to this. Do you have any sort of timetable for when we could expect the enhancement to be implemented?

wobba:
As I know Graph Search API is not hitting GCC High right away I won't do it right away, but within a couple of months should be doable as it shouldn't be too hard to get in. I'll likely add it to the last page of the settings pane defaulting to ww, and allow an override.

PR #1540, opened 2021-11-17
Search data sources updates

FranckyC:
- Updated Microsoft Search source to use the v1.0 Graph endpoint and let user decide to use beta as well.
- Added dynamic managed properties list for sort order in Microsoft Search source.
- Normalized refinement string generation for both sources.


wobba:
Can you consider adding a field to specify the Graph host? Related to #1246? As special clouds are not using graph.microsoft.com. Should be a collapsed/advanced setting.

FranckyC:
Good point @wobba, I'll add a new 'MS Graph' host then

PR #1561, opened 2021-11-25
Remove ID strings from highlighted summaries

phoob:
## Problem

the getSummary handlebar helper did not filter out ID strings from hitHighlightedSummary, resulting in summaries low legibility. Like this:

image

## Solution

Added a regex to remove all occurances of /\w+-\w+-\w+-\w+-\w+/ to the getSummary helper.

### DIFF
--- a/search-parts/src/services/templateService/TemplateService.ts
+++ b/search-parts/src/services/templateService/TemplateService.ts
@@ -428,7 +428,7 @@ export class TemplateService implements ITemplateService {
         // <p>{{getSummary HitHighlightedSummary}}</p>
         this.Handlebars.registerHelper("getSummary", (hitHighlightedSummary: string) => {
             if (!isEmpty(hitHighlightedSummary)) {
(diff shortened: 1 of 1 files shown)


wobba:
I'll think quickly about this and it's a well known problem even internally which we want to fix. The real issue is how SharePoint persists their pages, leading to "guids" in the indexed data. Which is then trickling all the way up to summaries. As the likelyhood of wanting a real guid showing is slim, we'll probably merge this one.

FranckyC:
Thanks @phoob !

Namrata99sah:
I have deployed the latest 4.5.3.0 PNP version to resolve this particular issue, but what I can see is the GUIDs are removed from most of the pages but for few pages it's still there. And not the full GUID, only a substring is present. I found it is happening for Quick Link webpart ID which is "c70391ea-0b10-4ee9-b2b4-006d3fcad0cd", in my search results its coming as -  "b2b4-006d3fcad0cd" or "4ee9-b2b4-006d3fcad0cd". So it's substring of the complete GUID. Pls help resolve this one.
@wobba @FranckyC @phoob 

image

wobba:
You can add your own replacement function to clean up the date... as it's a data issue coming from search, not a web part issue. And over the coming months the issue should be solved for newly saved pages as SharePoint is changing how modern pages are saved to avoid this issue.

PR #1578, opened 2021-12-01
Extend getPropertyValue return value by itemCount 

edarroudi:
as discussed in https://github.com/microsoft-search/pnp-modern-search/discussions/1577 i added the count of items to the getPropertyValue "dynamic data interface".

Tested the call and it worked well.


FranckyC:
Sounds good, thanks @edarroudi. We are looking forward to see what you gonna do with this ;)

Issue #1562, opened 2021-11-25
Page Lock (Conflict) with Multiple Results Web Parts


dfrancoeur:
**Version used**
4.3 and 4.4

**Description**
We have been using the web parts and running into a repeatable issue (across separate tenants) where the Search Web Parts seem to be breaking/interrupting the ootb Page Save behaviour. In monitoring via Fiddler I can see the SavePageAsDraft is still getting launched successfully (Page is getting saved), but the request that typically follows to reload the page never gets triggered. This leaves the page in a limbo state. Typically the page still has the UI of being in Edit mode (see the zones, editing buttons, etc.). Clicking on Save again yields the "Could not save your changes because a member has ended your session". 

I gradually simplified the page by removing web parts and found that I could eliminate this behaviour by reducing the Search Results web parts down to a single instance on the page. Has anyone else seen this behaviour?

Each vertical is setup to link to a specific Search Result web part - as per its primary usage pattern, see below for web part set up.

**Web Parts:**
- Search Box
- Search Vertical
- Search Refiner
- Search Results (1)
- Search Results (2)
- Search Results (3)
- Search Results (4)
- Search Results (5)
- Search Results (6)
- Search Results (7)

**Expected behavior**
Once the Page Save is clicked, the page should reload itself correctly out of Edit mode.

**Screenshots**

Example of page in Limbo State
![image](https://user-images.githubusercontent.com/79175670/143493435-98067801-4123-4c46-b3b5-f893f2d884d9.png)

Example of Could Not Save Changes
![image](https://user-images.githubusercontent.com/79175670/143493480-531f7bb6-9131-4963-a71d-3d19bde29d62.png)

**Desktop (please complete the following information):**
 - Edge Chromium

droper-lcb:
I actually saw this a few times myself yesterday. 
I've been making lots of small changes to a search result template, and after doing that for a while it would eventually have this happen, and then I refresh the page and its still in edit mode under my user and I would just have to get back to the webpart edit screens and continue on.

I thought it was actually another user kicking me out as we were both working on search templates throughout the day, but I guess not, as it wouldn't have kept it in edit mode to my user if that was the case.

I have a very similar screen layout as you describe, with multiple search results parts linked to show when the related vertical is selected.

till-llit:
i can reproduce this issue in our results page with multiple search result webparts (we are using search verticals).

edarroudi:
I can reproduce this behaviour as well.

[...]

PR #1591, opened 2021-12-03
Fixed `window.history.pushState` updated twice causing default SharePoint links to break

FranckyC:
Fixed an issue with page hyperlinks once the Search box WP and Search Results are loaded on the same page and the search box uses query string parameter or search query. window.history.pushShate method is update twice causing default SharePoint to break.


wobba:
Could be related to #1562

PR #1608, opened 2021-12-08
Vertical ResultCount Option

edarroudi:
Hi,

i created the following sampel code. I think i have to refactor it so it's just a draft / discussion basis.
Added Property "Show results count" to PnP Search Verticals.
image

if on, i start listening to the dynamicDataProvider.registerAvailableSourcesChanged

If a webpart source has a property "totalCount" in getPropertyValueAsync(source?.metadata?.alias), i accept it as result providing webpart and start listening to this.props.dynamicDataProvider.registerSourceChanged and this.props.dynamicDataProvider.registerPropertyChanged for that webpart.
So we are not only reducing the interface to "pnpSearchResultsWebPart", but accept other result webparts as well!

Then we show the sum of all counts of the connected webparts in the vertical pivot item:
image

if there are no results i don't show "(0)".
image

I have to add some checks for some scenarious but it seems to work fine just now. Any comments?

### DIFF
--- a/search-parts/src/models/dynamicData/IDataResultSourceData.ts
+++ b/search-parts/src/models/dynamicData/IDataResultSourceData.ts
@@ -26,4 +26,9 @@ export interface IDataResultSourceData {
      * The count of items returned by the getItemCount method of a datasource
      */
     totalCount?:number;
+
+    /**
+     * Passing the selectedVerticalKeys Properties
+     */
+     selectedVerticalKeys:string[];
 }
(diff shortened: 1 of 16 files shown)


FranckyC:
Hi @edarroudi, this one is pretty tricky. Since you can configure multiple Search Results Web Parts for a specific tab, you can't really assume a 1:1 connection. In this case, you should sum all counts to get a the correct count? Seems odd. An other big problem is you can only get the count for Web Parts currently displayed by the selected tab, not all tabs. The philosophy behind verticals is also to improve perfornances when several Web Part instances are on the page. When a Search Results Web Part is hidden (i.e. not associated to the current selected tab), no data are retrieved meaning no API call. Getting the count for this scenario would mean to execute the query for every single Web Parts, which is a nonsense regarding the initial wanted behavior. FYI that was the approach took by the v3 and that was not very flexible. Due to the current architecture, I don't have a clean solution for now and I couldn't merge what you are trying to do for that reasons.

edarroudi:
Hey @FranckyC ,

thanks for the detailed reply. I agree with you - it just would have worked for my project :)

FranckyC:
I close this one as I couldn't be merged as is.

PR #1609, opened 2021-12-08
Bugfix: "useVerticals" and not "useFilters" property

edarroudi:
Small fix


FranckyC:
Good catch. Thanks @edarroudi

PR #1610, opened 2021-12-08
Fix duplicities in proposed suggestions

cebud:
In case, when you are writing slowly search input value, in combination with slow async suggestion provider, method updateQuerySuggestions is triggered multiple times (previous execution is not yet finished). Result is that suggestions are duplicated. 

Proposed solution will wait for condition: inputValue is the same as searchInputValue and skip the state, when termToSuggestFrom is empty. I don't understand exactly, why condition !termToSuggestFrom has been in line 168.


FranckyC:
Thanks @cebud !

PR #1624, opened 2021-12-13
Ensure URL hash is always at the end of the string

stigre:
Currently there is a bug in the filters webpart when the search box is connected to the URL fragment. The search terms are always reflected in the URL hash, but when filters are applied, the deep link added to the URL does not respect the hash, and is instead appended after it. This results in the entire filters deep link being considered part of the search terms, and it breaks your search.

image

This PR ensures that the hash is always at the end of the new URL when filters are added/removed.


FranckyC:
Hi @stigre, good catch! Thanks

Issue #905, opened 2021-03-31
The field value doesn't appear when the fieldname contains "-"


hervedorierai3:
**Version used**
v4

**Describe the bug**
If the site columns name contain a '-', the {{slot item @root.slots.Field2}} directive doesn't show the field value.

**To Reproduce**
1. Create a site column named "FieldTest" (Number)
2. Create a site column named "Field-Test" (Number)
![bug-1-sitecolumns](https://user-images.githubusercontent.com/68116737/113211443-c2164380-9275-11eb-89e9-63d4269507df.png)
2. Create a content type with these site columns (Element).
3. Create a list, and add the content type.
4. Add content in the list.
5. Create a page and the Search Result webpart.
6. Configure the webpart to search in the list.
7. Click on the "Selected properties" button to select the two fields
![bug-2-selectfield](https://user-images.githubusercontent.com/68116737/113211501-d35f5000-9275-11eb-93d8-ab457772f671.png)
8. Click on the "Customize" button in the "Layout slot" section and add two new slots : 
![bug-3-slots](https://user-images.githubusercontent.com/68116737/113211531-dce8b800-9275-11eb-907b-5ee1d62e2a52.png)
9. Select the "Custom" layout.
10. Edit the template to add : Field-Test={{slot item @root.slots.Field1}} FieldTest={{slot item @root.slots.Field2}}

**Expected behavior**
The values of each field ("FieldTest" and "Field-Test") should be displayed but, only the "FieldTest" value is displayed.

**Screenshots**
![bug-4-result](https://user-images.githubusercontent.com/68116737/113211549-e40fc600-9275-11eb-87dc-f8d460aeffc1.png)

**Desktop (please complete the following information):**
 - Edge

**Additional context**
N/A

FranckyC:
Hi @hervedorierai3, as a workaround, you can still use a `RefnableStringXXX` managed property mapped to your field crawled property. Having a '-' in a field name is not really a best practice. Do you have any error message in the browser console or network tab?

hervedorierai3:
Hi @FranckyC.
Thank you for your feedback... I did the best workaround : I recreated the field without "-" ! ;-) I know it's not a good practice, but SharePoint allows to do that. A lot of customers creates site columns with space, or ', é, ... 
So I'm not blocked by this issue, but I think it's a good idea to fix it.

ayblanchet:
Hi, @FranckyC @hervedorierai3, 

[...]

PR #1673, opened 2021-12-23
Updated ObjectHelper byPath

FranckyC:
Updated the ObjectHelper,byPath method to handle wrong expression correctly


wobba:
@FranckyC did you verify the code change with the issue in https://github.com/microsoft-search/pnp-modern-search/issues/905?

PR #1686, opened 2022-01-04
Option to override image size in pnp-img

phoob:
## Problem

Needed a way to override image size in pnp-img

## Solution

Extend the pnp-img tag with a custom property (data-image-size), and replace /large/ in AutoPreviewImageUrl if this exists.


wobba:
Can you add a hyperlink reference to docs explaining the `c194x110_crop` notation? And also note that this only works for these special thumbnail size url's, not for a random image.

Or you rework it so that passing in something like `c194x110_crop` adds the required CSS to scale/crop an image id addition.. then it will work for any image URL.

phoob:
I now realize that the AutoPreviewImageUrl is generated by pnp-modern-search (in SearchResultsContainer) – so there is probably a better way of achieving this...

PR #1710, opened 2022-01-12
Fix selected filter tags from overflowing

mbice:
Selected filter tags that are really long will overflow the container. This PR fixes that so they don't exceed the size of the container and the ellipsis overflow is correctly applied.

So it goes from this:

image

To this:

image

Thanks!


wobba:
Thank you :)

Issue #1623, opened 2021-12-13
Custom Template Paging Style issue ( Style cleared after paging event)


edarroudi:
**Version used**
4.5.3

**Describe the bug**
Custom styles inserted in an custom template are not rendered after a paging event.

**To Reproduce**
1. Add SearchBoxResult Webpart
2. Select SharePoint Provider
3. Select ResultSource LocalPeopleResults
3. Select People Template
4. edit result template 
5. add custom style .ms-Persona-tertiaryText
        {
            font-weight:600;
        }

![image](https://user-images.githubusercontent.com/935493/145799574-583375fd-e393-4583-ac08-4944701fb0c9.png)
6. Save template
7. Save page
8. Page the site

**Expected behavior**
Style tag shouldn't be empty

**Screenshots**
![image](https://user-images.githubusercontent.com/935493/145800296-2b6e147e-b74b-4085-bf92-b1467bab7de1.png)
![image](https://user-images.githubusercontent.com/935493/145800809-518752d9-32f8-4d2e-aac9-7bc720796eaa.png)

**Desktop (please complete the following information):**
 - Browser edge, chrome

**Additional context**
Also: it seems to not be a problem of paging, but also if you do a second query. Like search something and it the style is filled / rendered correctly. Then without a page refresh, search for something else => style is not rendered

**Update**: 
Yes, only happens in People template.
Additionally: if you have a results webpart with people template on a page and customized other results webparts, the people template will break the other templates as well.

edarroudi:
I added a PR with a "Fix" for it https://github.com/microsoft-search/pnp-modern-search/pull/1730
Funny stuff.

FranckyC:
Fixed with #1730

PR #1730, opened 2022-01-18
Removed forbid_tags (style) from dompurify - PersonaComponent

edarroudi:
Issue mentioned in https://github.com/microsoft-search/pnp-modern-search/issues/1623
If PersonaComponent is used, domPurify removes the style component in the template after first paging event or new search.

After removal of the FORBID_TAGS:['style'] config line the issue is fixed.

It still is a mystery to me why this bug occurs. I think the bug iteself is on a higher level, so actually what you guys want is to purify everything that the component renders but it actually purifies the html code surrounding the component as well.

Probably "someone ;-)" needs to check if this works as you guys intended because the generel usage of dompurify is equal in all components - so it works but actually does more than it should?

### DIFF
--- a/search-parts/src/components/PersonaComponent.tsx
+++ b/search-parts/src/components/PersonaComponent.tsx
@@ -85,7 +85,6 @@ export class PersonaComponent extends React.Component<IPersonaComponentProps, IP
         this._domPurify = DOMPurify.default;
 
         this._domPurify.setConfig({
-            FORBID_TAGS: ['style'],
             WHOLE_DOCUMENT: true
         });
 


wobba:
I'll have to think about this one a bit :)

edarroudi:
Hey @wobba ,
i checked the code know. And obviously that wasn't the correct solution to the problem.

What happens is actually:

**First Render:**
    TemplateRenderer.tsx in constructor: 
```
setConfig( {
          ADD_TAGS: ['style'],
         ADD_ATTR: ['target', 'loading'],
         ALLOW_DATA_ATTR: true,
         ALLOWED_URI_REGEXP: /^(?:(?:(?:f|ht)tps?|mailto|file|tel|callto|cid|xmpp|xxx|ms-\w+):|[^a-z]|[a-z+.\-]+(?:[^a-z+.\-:]|$))/i,
         WHOLE_DOCUMENT: true,
    })
```
updateTemplate is called, purifies with the correct config from above
    PersonaComponent in constructor: 
```
    this._domPurify.setConfig({
            FORBID_TAGS: ['style'],
            WHOLE_DOCUMENT: true
        });
```

Persona calls sanitize with the correct config for itself.

First Render works

**Second Render**
       updateTemplate is called, but this time with the config from PersonaComponent :(
       Persona (correct :)

Probably have to pass the config everytime to the sanitize method for every call everywhere?

wobba:
We should probably use the same config everywhere. Not sure why persona has forbid on style. I think we can remove it, similar to the other components. I think that's the fix.

FranckyC:
@wobba give me some time to check why I did this before processing. Unfortunately quite busy right now :/

FranckyC:
@edarroudi  @wobba I actually don't recall why I didn't not allow styles on persona card :/. Removing it should be OK as it is the same as cards. Did some tests and no problem.

PR #1734, opened 2022-01-19
https://github.com/microsoft-search/pnp-modern-search/discussions/1688

wuxiaojun514:
https://github.com/microsoft-search/pnp-modern-search/discussions/1688

I created a new property for "Hit Highlighted Properties" in PnP Search Result Webpart
The usage is the same as the "SharePoint Search Query Tool"
image

You need add "HitHighlightedProperties" in "Select Property" when you want to use it
image

Then you will see it the result like '"_**HitHighlightedProperties": "<HHTitle>Wu, Xiao Jun <c0>Tony</c0></HHTitle><HHUrl></HHUrl><username hashh=\"1\"><c0>Tony</c0>.Wu@dnvdev.onmicrosoft.com</username>",**_
image

I updated the language resource files by bing translator, maybe we need update them with real translator


wobba:
Can you please check the correct translations using the classic result web part?
image

I want you to move the section to the bottom as it's a special use case and should not be prominent.

You also need to update the doc pages to reference the change.

wuxiaojun514:
Hello @wobba , I will compare the text on classic sharepoint search result webpart

I moved the position of "Hit Highlighted Properties" after "enable localiation"
image

I will update the related doc page and back to you

wuxiaojun514:
Hello @wobba 

- I updated the language resource files to use the text from classic search result webpart (It is server side webpart, so I have to switched language, got text ; switched language, got text...)

- The Hit Highlighted Properties has been moved to the bottom (after enable localization)

image

- I updated the document file to append Hit Highlighted Properties on it
image

FranckyC:
Merged with #1790. Thanks @wuxiaojun514.

PR #1786, opened 2022-02-01
Feature/fix query alteration options

edarroudi:
The query alteration structure is different on request and response.
See https://docs.microsoft.com/de-de/graph/search-concept-speller
The options have to be inside the "request" structure, not outside.
The queryAlterationResponse object is inside the "value" response.

I don't know if i/we did some mistake or Microsoft moved the structure inside the request. The updated article date suggests that something did change. Perhaps @wobba knows more about it. 
image


edarroudi:
Saw this interesting piece as well.. 
image

Shall i change how we handle the queryTemplate as well or is it too soon? Currently:
image

FranckyC:
Hi @edarroudi , `queryAlterationResponse` is still only available in beta so we won't pocess it until it moves ot v1.0 endpoint. Thanks!

edarroudi:
Hi, it's not beta anymore: https://developer.microsoft.com/en-us/graph/changelog/?page=2
I modified it (no requirement for beta flag) and updated the documentation.
image

PR #1813, opened 2022-02-09
add new scenario 9

davemehr:
I have created the new pnp modern search scenario 9. Add images in assets. Updated the index with the new scenario.


davemehr:
Yes I can, and all other entries move up a number? Or does we want to continue without "number"?

wobba:
I think removing the numbers is the way to go.. makes it easier :)

davemehr:
Yes, agreed ;) . I will rearrange the scenarios without numbers.

Issue #972, opened 2021-04-29
PnP search Refiner WebPart(v4.1) – File Types should group and show user friendly names


ShariqSiddiquie:
**Is your feature request related to a problem? Please describe.**
Search refiner template group the similar file types and show user friendly names.
We have this feature in SharePoint 2013

**Describe the solution you'd like**
All similar file types should club in one group and the group name should be user friendly.
Ex :- “Word” group (.doc, .docx)

**Describe alternatives you've considered**
This feature should be in PnP Search refiner similar as SharePoint 2013. Actual file type names are not user friendly.   

**Additional context**
Please see the attached document with screenshots

[PnPSearchRefiner(v.1) FileType group and user friendly names.docx](https://github.com/microsoft-search/pnp-modern-search/files/6399001/PnPSearchRefiner.v.1.FileType.group.and.user.friendly.names.docx)

ShariqSiddiquie:
This is a very good feature in SharePoint 2013. It would be great if we get this feature in PnP Refiners. 
When we can expect to get this feature in PnP Search Refiners?

wobba:
You would need one or more property matching expressions which is used to group over to a UX label. Pseudo logic below, and question is how advanced you want to get with AND/OR's.

```
Label = Document
   FileType contains doc
   OR FileType contains pdf
   AND SPContentType equals CoolDoc
Label = Spreadsheet
   FileType contains xls
```

where the expressions must support being translated into FQL when you click on them.

edarroudi:
Hey @wobba ,
i'll start with  this now - just a general question: I would like to use Hooks instead of the old class centric React development. Would that be fine with you guys?

wobba:
@edarroudi You can use hooks if you want, but mixing and matching patterns may make code more cluttered. So I'm a bit ambivalent.

FranckyC:
Same here, I find hooks hard to read in my opinion. If you prefer this option, why not but make it easy to read and follow.

edarroudi:
Hi guys, 
i implemented a bit of the layout for the grouped refiners and wanted to ask if i go into the right direction here.
I believe that most users only need a simple grouping with a list of values that belong to the "group label".
I have an advanced option in mind where the user can just write some FQL themselves - with no checks on our end.
Do you see any issues here? I don't want to implement a FQL editor or funny stuff like that. Mayvbe we have to think about some security issues here.

Another question would be the sorting behaviour: 
[...]

PR #1845, opened 2022-02-17
972: Refiner Grouping 

edarroudi:
See https://github.com/microsoft-search/pnp-modern-search/issues/972
This is just a draft for now as there might be the need for additional options, security checks ...
image
image
image
image
image

Todos:

1. Additional language resources
2. Check for possible security issues
3. Update documentation
4. Many tests / combinations
3. How to go on with the counts. Its probably impossible to correctly calculate the "advanced" mode count

**Decided to check the following 2 options in another version. Will add too much complexity right now.**
 _[optional] Allow the usage of Tokens in the advanced (and simple) fql.
 [optional] Allow user to set if Grouped labels should stay of the top of the Combobox/Checkbox_

Regarding the advanced FQL request: It works but i'm not sure if it is really correct from the syntax :)


wobba:
Sorry we don't have time to properly test this for the next release, but will get it included in the v4.8 release.

wobba:
Should we set this from draft and start to address it?

edarroudi:
> Should we set this from draft and start to address it?

no I still have to adress some issues / optimize it. I'm thinking of switching to a rule tree control for the advanced KQL mode but I have to finish that controll here https://github.com/pnp/sp-dev-fx-property-controls/pull/456 (very soon).

Will concentrate on this PR beginning next week. Do you know when you guys will publish 4.8.0 ?

edarroudi:
@wobba I had some time to think about it. The advanced Editor is not necessary and as it would lead to a lot of work I think we should move forward with this solution. My questions / Todos are still open - I will continue / finish the implementation once you and @FranckyC come to a conclusion or have some remarks on this - otherwise I will be investing a lot of time without knowing if you guys are actually ok with the approach. Thats why I set this now to Ready for reveiew - although it's not really 100% ready ;-)

edarroudi:
Update 2022-11-11
I'm currently updating the handling of Grouped Filter (Labels) and testing everything. Will request a review once done (probably by monday).

edarroudi:
I fixed some issues and tested it many times. From my perspektive it works fine. 
I don't allow refiner grouping on date type templates yet. Maybe in a future version? Don't know if i would have to check several issues there.

@FranckyC do you have some time to check / comment this?

Once I get a go I update the rest of the language files and the docs. **Thank you.**

edarroudi:
I will add and test number fields with ranges next week. And check what has to be done for date fields

edarroudi:
For now again in draft, will recheck this. Maybe wil start with some refactoring of the base filter code itself in another branch.

wobba:
Should we close this one?

edarroudi:
Yeah there is too much work left for it to work in every scenario. And those users who want to have grouped refiners can implement it through UI in the current version.

mozilla0:
@edarroudi where did you implement this in the UI?

MirasMNC:
> Yeah there is too much work left for it to work in every scenario. And those users who want to have grouped refiners can implement it through UI in the current version.

Hi, may I ask you how is this possible to do it via UI? Currently we are facing the same issue that end users gave us feedback about filters. It would be nice to have an option to group some filetypes and give them friendly display name. Thank you :)

wobba:
@MirasMNC someone needs to invest developer time to create a filetype filter which works semantically and on the pure values.

Issue #1372, opened 2021-10-06
Localized dates not available?


phoob:
I have been trying to display dates in a search results template, but I am not able to get the correct localization. 

The UI language is set to Norwegian for me, so I would expect
```{{getDate RefinableDate00 "LL"}}```
to show `6. oktober 2021` but it shows `October 6, 2021`

According to the docs, I should also be able to use {{moment}}, where I can set the "lang" attribute. So I would also expect
```{{moment RefinableDate00 lang="nb" format="LL"}}```
to show `6. oktober 2021`, but that gives me the the same result as above (`October 6, 2021`). I have also tried
```{{moment RefinableDate00 locale="nb" format="LL"}}```.

Using `{{moment}}` I have also tried other locales, such as "fr" and "fi", but the dates are still displayed in English.

Please help :)

**Comment by phoob** [comment]
> Duplicate of #1253. Can you try with the `develop` branch?

Thanks! It works using `{{moment}}` in develop.

I still get English when I use `{{getDate}}`, but curiously it prints the date in Norwegian when I use it after a moment tag… so:

```
  <p>{{getDate (slot item @root.slots.StartDate) "LL"}}</p>
  <p>{{moment (slot item @root.slots.StartDate) lang="nb" format="LL"}}</p>
```
prints
```
October 4, 2021
4. oktober 2021
```
but 
```
  <p>{{moment (slot item @root.slots.StartDate) lang="nb" format="LL"}}</p>
  <p>{{getDate (slot item @root.slots.StartDate) "LL"}}</p>
```
prints
```
4. oktober 2021
4. oktober 2021
```

**Comment by phoob** [comment]
Ok.. tested a bit more. It seems it is because of the moment tag I used before the second getDate... if I change the line

```
            {{#if (moment (slot item @root.slots.StartDate) isAfter=(moment format="YYYY-MM-DD"))}}
```

to 

```
            {{#if (moment (slot item @root.slots.StartDate) lang="nb" isAfter=(moment lang="nb" format="YYYY-MM-DD"))}}
```

the subsequent getDate yields Norwegian formatting…

eduardpaul:
I'm pretty sure that it's the same issue that I had with Spanish: SharePoint UI Culture is "es-es" but momentjs has only "es.js" locale file. After the change: 

![image](https://user-images.githubusercontent.com/1370426/161131854-b2a72405-9cd2-448e-99dd-84eb20a111ca.png)

[...]

PR #1986, opened 2022-03-31
Fix momentjs locale

eduardpaul:
- Momentjs locale files now included in production bundle.
- Fixed several languages in momentjs that are using two letter file name (nb-NO, es-es,...) instead SP ISO format. 
- This PR should close [this](https://github.com/microsoft-search/pnp-modern-search/issues/1372) as it's working with 'es-es'.

### DIFF
--- a/search-parts/gulpfile.js
+++ b/search-parts/gulpfile.js
@@ -76,10 +76,12 @@ const envCheck = build.subTask('environmentCheck', (gulp, config, done) => {
                 }
             });
 
-            // Exclude moment.js locale for performance purpose
-            generatedConfiguration.plugins.push(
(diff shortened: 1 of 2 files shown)


FranckyC:
Hi @eduardpaul, thanks for the fix. It actually works be it also seems some locales are missing from your array. For instance, I had to add 'fr' manually in the `momentTwoLetterLanguageName` array. Can you do a quick check to ensure all supported languages (i.e. the ones who have resources files in the solution) are at least present? Thanks!

eduardpaul:
Very good point @FranckyC  fr-fr and nl-nl was missing. Didn't add them because momentjs has fr-ca and fr-cha and I didn't realize that the least exotic was missing! :D

FranckyC:
That was quick, thanks again @eduardpaul !

PR #1541, opened 2021-11-18
Create de-de.js / German language

dominiqueborer:
Added German to the repo

PR #1987, opened 2022-03-31
Updated: Create de-de.js / German language based on 1541

edarroudi:
Updated original de-de.js file from [dominiqueborer](https://github.com/dominiqueborer) in #1541 
Added the german webpart loc files


FranckyC:
Danke @edarroudi !

Issue #2059, opened 2022-05-03
Use anchors in search verticals 


d-martins:
**Is your feature request related to a problem? Please describe.**
When a search vertical item is set as hyperlink, the rendered DOM does not include anchor elements to handle the navigation. As a result, the soft-navigation from SharePoint modern pages is not triggered and the browser needs to fully reload the clicked vertical's page.

**Describe the solution you'd like**
If a vertical is set to be an hyperlink, SharePoint's soft navigation should be triggerable

**Additional context**
These videos compare the current navigation (button click event) and possible anchor navigation.
The anchors are being injected by a script in the console, replacing the text span with an anchor using mutation observers.

https://user-images.githubusercontent.com/6013384/166473596-98706130-3ccb-4820-9cf4-e29e61308f72.mp4

https://user-images.githubusercontent.com/6013384/166473600-5c1be9a8-d275-42d5-a701-282e5682edd6.mp4

PR #2065, opened 2022-05-04
Support SharePoint soft navigation

d-martins:
See [#2059](https://github.com/microsoft-search/pnp-modern-search/issues/2059)

**To test the solution:**
- Create 2 pages in a SharePoint site collection
- Add the PnP -Search Verticals web part to each
- In one page, configure 2 verticals, one for each page
  - The verticals should be hyperlinks
  - The hyperlink should be configured to open on the current tab
- Copy the settings from the web part and apply them on the second page
- Publish both pages
- Click on the tabs. Only the content area should refresh

### DIFF
--- a/search-parts/src/webparts/searchVerticals/components/SearchVerticalsContainer.tsx
+++ b/search-parts/src/webparts/searchVerticals/components/SearchVerticalsContainer.tsx
@@ -95,7 +95,17 @@ export default class SearchVerticalsContainer extends React.Component<ISearchVer
           const behavior = vertical.openBehavior === PageOpenBehavior.NewTab ? '_blank' : '_self';
           this.props.tokenService.resolveTokens(vertical.linkUrl).then((resolvedUrl: string) => {           
             resolvedUrl = resolvedUrl.replace(/\{searchTerms\}|\{SearchBoxQuery\}/gi, GlobalSettings.getValue(BuiltinTokenNames.inputQueryText));
-            window.open(resolvedUrl, behavior);
+
+            if(behavior === "_blank"){
+              window.open(resolvedUrl, behavior);
+            }else{
+              // Allow SharePoint to intercept the click and do a soft navigation
+              document.body.insertAdjacentHTML('beforeend', `<a target="${behavior}" href="${resolvedUrl}" style="display:none;"></a>`);
+              const anchor = document.body.lastElementChild as HTMLElement; 
+              anchor.click();
+              document.body.removeChild(anchor);
+            }
+
           });
           
       } else {
--- a/search-parts/src/webparts/searchVerticals/components/SearchVerticalsContainer.tsx
+++ b/search-parts/src/webparts/searchVerticals/components/SearchVerticalsContainer.tsx
@@ -92,15 +92,14 @@ export default class SearchVerticalsContainer extends React.Component<ISearchVer
       const vertical = this.props.verticals[verticalIdx];
       if (vertical.isLink) {
           // Send the query to the new page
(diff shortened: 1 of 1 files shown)


search-parts/src/webparts/searchVerticals/components/SearchVerticalsContainer.tsx
```diff
@@ -95,7 +95,17 @@ export default class SearchVerticalsContainer extends React.Component<ISearchVer
           const behavior = vertical.openBehavior === PageOpenBehavior.NewTab ? '_blank' : '_self';
           this.props.tokenService.resolveTokens(vertical.linkUrl).then((resolvedUrl: string) => {           
             resolvedUrl = resolvedUrl.replace(/\{searchTerms\}|\{SearchBoxQuery\}/gi, GlobalSettings.getValue(BuiltinTokenNames.inputQueryText));
-            window.open(resolvedUrl, behavior);
+
+            if(behavior === "_blank"){
```
wobba, line 99: I suggest to remove L95 and do the check on `vertical.openBehavior === PageOpenBehavior.NewTab` here

search-parts/src/webparts/searchVerticals/components/SearchVerticalsContainer.tsx
```diff
@@ -95,7 +95,17 @@ export default class SearchVerticalsContainer extends React.Component<ISearchVer
...
+              window.open(resolvedUrl, behavior);
```
wobba, line 100: Use `_blank` here

search-parts/src/webparts/searchVerticals/components/SearchVerticalsContainer.tsx
```diff
@@ -95,7 +95,17 @@ export default class SearchVerticalsContainer extends React.Component<ISearchVer
...
+            }else{
+              // Allow SharePoint to intercept the click and do a soft navigation
+              document.body.insertAdjacentHTML('beforeend', `<a target="${behavior}" href="${resolvedUrl}" style="display:none;"></a>`);
```
wobba, line 103: And you can remove the target attribute here.

search-parts/src/webparts/searchVerticals/components/SearchVerticalsContainer.tsx
```diff
@@ -92,10 +92,19 @@ export default class SearchVerticalsContainer extends React.Component<ISearchVer
...
-            window.open(resolvedUrl, behavior);
+
+            if(vertical.openBehavior === PageOpenBehavior.NewTab){
+              window.open(resolvedUrl, "_blank");
+            }else{
+              // Allow SharePoint to intercept the click and do a soft navigation
+              document.body.insertAdjacentHTML('beforeend', `<a href="${resolvedUrl}" style="display:none;"></a>`);
+              const anchor = document.body.lastElementChild as HTMLElement; 
+              anchor.click();
+              document.body.removeChild(anchor);
```
wobba, line 105: Any chance of getting an error here if the navigation happens before the DOM is removed? Adding a try/catch around removeChild ensures that this case never happens.

d-martins: This is unlikely to happen, as the code block should run synchronously before the click event is resolved.
But in any case, we can validate if the anchor.parentElement is not null before removing, which would be a valid operation even if it weren't in the DOM somehow, and wouldn't result in an error.

```suggestion
              anchor.parentElement?.removeChild(anchor);
```

wobba:
Thank you :)

PR #2090, opened 2022-05-16
Checking nulls when clearing search

rick-brown-slalom:


wobba:
Hi, looks ok, but do you have an issue to link this to? Would be nice to know the scenario where it breaks.

rick-brown-slalom:
We had a lot of web parts on the page and I haven't tested this with a clean page, but basically the currentTarget didn't exist when pressing the 'X' on the Search box, and therefore it failed to clear the search box of our search term.

wobba:
Thank you :)

PR #2108, opened 2022-05-24
Specify fix

mrjarbenne:
Noticed that specify had been spelled incorrectly in a few different places. This Pull should fix it.


FranckyC:
Hi @mrjarbenne thanks for the fix ;)

Issue #2072, opened 2022-05-08
Filter issue with localization enabled


coupalm:
**Version used**
Ex: 4.6.1.0

**Describe the bug**
Filter does not always correctly apply on search result for a managed property using localization.

Search result is using localization. I have a RefinableString04 mapped to a managed metadata column ows_taxId. In the template I'm using the AutoRefinableString04 managed properties. In th search results terms are correctly translated.
However, in the search filter, I'm using RefinableString04 as a filter, the translated value correctly displays but when I apply the filter, the search results are not always correctly filtered.

**To Reproduce**

Use a term set with translated value
Add a managed metadata column mapped to this term set to a library
Add some documents and set with different terms
Use a RefinableString managed property mapped to the ows_taxId of the added column.
Create a page and add pnp search result web part. Set query to return only documents for the libary url. Add the newly created managed property to the selected properties. Enable localization
Using the detail list layout, add a column using expression {{AutoRefinableStringXX}}
Should see the value of RefinableStringXX correctly translated.

Add a PNP search filter
Add the RefinableStringXX as a filter. Link to the search results web part. Do the same thing with the search result web part (specify it is using this filter web part).
Values in the filter web part will display translated value of RefinableStringXX correctly.

However, using a value in the filter will not always correctly filters the results.

**Expected behavior**
Only results with the selected value should be returned

**Screenshots**
![image](https://user-images.githubusercontent.com/11309224/167311785-bfcc5c30-0cbd-4638-ae4b-f6b2b3e91ebd.png)
![image](https://user-images.githubusercontent.com/11309224/167311793-3acb0913-c515-4183-b36c-c3061489aa71.png)

**Desktop (please complete the following information):**
 - Browser Edge latest release

wobba:
Do you have a sample request JSON for the request made you can share? That helps troubleshooting this.

coupalm:
Here are more screenshots.

No filter applied:
![image](https://user-images.githubusercontent.com/11309224/167412059-bbf53b20-1e8c-4a9d-ae5b-c97ddfdf7b83.png)

Select first term in the filter list:
![image](https://user-images.githubusercontent.com/11309224/167412177-e705c8ef-d362-40c5-9a7f-88bc3d126edd.png)

Select second term in the filter
![image](https://user-images.githubusercontent.com/11309224/167412311-c6f7638f-bfe3-442e-be2f-ae00e4ebdf4a.png)

Select third term
[...]

PR #2110, opened 2022-05-25
WIP: Fix of localized refinement issue

eduardpaul:
We are using the following change in refinement to fix the https://github.com/microsoft-search/pnp-modern-search/issues/2072 issue. "Works on my machine" kind of PR, if anyone can test that it works in all the cases, it would be appreciated.

PD.: The credit of this PR goes to my colleague Martin Luis


FranckyC:
Hi @eduardpaul, thanks for the PR. Tested but does not work with this use case unfortunately https://microsoft-search.github.io/pnp-modern-search/usage/search-filters/#use-indexed-property-bag-properties-with-taxonomy-values.

FranckyC:
The combination that seems to work in every case seemst to be `or(${termFilterWithoutTranslations},${termTextFilter})` without the HEX encoding and skipping the original value with language specific label.

PR #2135, opened 2022-06-06
Added search options

FranckyC:
- Added a trim duplicates options for SharePoint search and Microsoft Search (beta only)
- Added support of query template in Microsoft Search (beta only)
- 
Discussion related #2088


wobba:
I think we should revert the trimduplicates feature and add collapsespec instead if we are going that route.

PR #2147, opened 2022-06-10
Update to SPFx v1.14.0

wobba:


FranckyC:
@wobba should I include it in the next beta release?

PR #2244, opened 2022-07-20
Scroll to view top of results

emilyk0520:
This PR restores a [fix made in v3](https://github.com/microsoft-search/pnp-modern-search/pull/709) of the Modern Search web parts.

When navigating through the pages of results, the view will scroll to the top of the search web part if not in the viewport.


FranckyC:
Thanks @emilyk0520 !

PR #2400, opened 2022-09-29
Enable custom data source extensibility

patrikhellgren:
Enables custom data source extensibility as discussed in #2397.
Also includes documentation for the feature.
A separate PR with example usage has been submitted to [https://github.com/microsoft-search/pnp-modern-search-extensibility-samples](https://github.com/microsoft-search/pnp-modern-search-extensibility-samples).

### DIFF
--- a/docs/extensibility/custom_data_sources.md
+++ b/docs/extensibility/custom_data_sources.md
@@ -1,0 +1,74 @@
+# Create a custom data source
+
+Custom data sources can be added to a search results Web Part to get results from your custom source.
+
+## Custom data source creation process
+
+Custom data source creation process comes in two distinct steps:
+
+1. [Create the data source logic](#create-the-component-logic-and-sub-components).
+2. [Register the data source information for discovery](#register-component-information).
+
+### Create the data source logic
+
+* In your extensibility library project, create a new `CustomDataSource.ts` TypeScript file.
+* Create an interface for your data source properties, typically the ones you want to persist in the Web Part property bag. Data source properties are isolated from the other general Web Part properties under the property `dataSourceProperties` in the property bag object.
+```typescript
+    export interface ICustomDataSourceProperties {
+        possibleResults: string;
+    }
+```
+
+* Implement the `BaseDataSource` abstract class using your properties interface:
+```typescript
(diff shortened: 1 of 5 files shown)


wobba:
Thank you, and we'll take a look and process these. Might take some weeks to set expectations due to time availability :)

FranckyC:
Hi @patrikhellgren, tested and works correctly. However, see my comment regarding the provided sample. A more advanced data source would be more useful to get started. Thanks!

wobba:
@FranckyC I think we can merge now with the existing sample, and add a more complex one later.

edarroudi:
I can add a generic REST DataSource sample later on. I already developed a multi datasource interface and could just extract the REST DS as a sample. 
image
I just would add the REST DS Interface as an interface 
image


I'll do it once 4.8 is released including the external datasource interface...

edarroudi:
@FranckyC / @wobba . I'm starting to create the extract of the above REST DataSource Sample for the samples. It could be finished in 1-2 days. Question: Would you be fine with that sample or do you have some concerns, e.g. Security issues?
Optionally I would have a switch to let users use the AAD/MSAL client (AadHttpClient) to authenticate to e.g. Azure Functions ...
Maybe a combobox: HttpClient| HttpClient with AadTokenProviderFactory (extra token url for bearer token configurable) | AadHttpClient | SPHttpClient 

I'll do the extraction anyways but it would be nice to hear a comment about any restrictions I should add beforehand.

edarroudi:
> My opinion is that a sample don't have to be to complex, as complexity should be handled by someone creating a provider for whatever system they use. Unless you want to showcase a good pattern with e.g. MSAL which is more work :)

Yeah my extended idea / extract would be probably to complex. But having a (minimal) configurable Rest DataSource could be good  starting point. So I would excludeMSAL integration and some attributes like httpheaders... and just allow the URL, Method and BODY Property. Thats just a couple lines of code and not too complex?!

edarroudi:
Created https://github.com/microsoft-search/pnp-modern-search-extensibility-samples/pull/15 . Personally I think @patrikhellgren  sample is more than enough.

PR #2446, opened 2022-10-14
Beta Entity Types Bookmark & Acronym

edarroudi:
Bookmark & Acronym work fine so far...

Bookmark & Acronym don't work together with Entity "DriveItem".
Some Properties have to be restricted when using Bookmark & Acronym, like "from" (Paging)... Those restrictions can be applied later?

Shouldn't we limit choosing those new Entity Types (Bookmark, Acronym, chatMessage...) to be only available if the "Use beta endpoint" switch is on? 

I would do this in another PR.

Althoug I don't know if those lists are up to date:

[1.0](https://learn.microsoft.com/en-us/graph/api/resources/search-api-overview?view=graph-rest-1.0&preserve-view=true#scope-search-based-on-entity-types)

[beta](https://learn.microsoft.com/en-us/graph/api/resources/search-api-overview?view=graph-rest-beta&preserve-view=true#scope-search-based-on-entity-types)

|Entity|1.0|Beta|
|-|-|-|
|message|x|x|
|event|x|x|
|drive|x|x|
|driveItem|x|x|
|list|x|x|
|listItem|x|x|
|site|x|x|
|externalItem|x|x|
|person||x|
|chatMessage||x|
|bookmark||x|
|acronym||x|

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -24,6 +24,10 @@
     | _Calendars.Read_ | Allow search for user's calendar appointments using Graph API (Events).
     | _Sites.Read.All_ | Allow search for sites using Graph API (Sites / List Items).
     | _ExternalItem.Read.All_ | Allow search for connector items using Graph API (External Items).
+    | _Bookmark.Read.All_ | Allow search for Bookmarks in Microsoft Search in your organization.
+    | _Acronym.Read.All_ | Allow search for Acronyms in Microsoft Search in your organization.
+    | _Chat.Read_ | Allow search for Teams messages.
+    | _ChannelMessage.Read.All_ | Same as above.
 
 3. Add the Web Parts to a SharePoint and start building!
 
--- a/search-parts/config/package-solution.json
+++ b/search-parts/config/package-solution.json
@@ -11,7 +11,7 @@
             {
                 "resource": "Microsoft Graph",
                 "scope": "Presence.Read.All"
-            },            
+            },
             {
                 "resource": "Microsoft Graph",
                 "scope": "User.Read"
@@ -55,6 +55,14 @@
             {
                 "resource": "Microsoft Graph",
                 "scope": "ChannelMessage.Read.All"
+            },
+            {
+                "resource": "Microsoft Graph",
+                "scope": "Bookmark.Read.All"
+            },
+            {
+                "resource": "Microsoft Graph",
+                "scope": "Acronym.Read.All"
             }
         ],
         "metadata": {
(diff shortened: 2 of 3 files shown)


wobba:
Thank you , and you should provide feedback to the Graph team on the documentation cases.

ZChristine:
hi, @edarroudi. Thanks for reaching out. 

Bookmark & Acronym don't work together with Entity "DriveItem". **yes. we do not support combination search for bookmark and acronym.** 

Some Properties have to be restricted when using Bookmark & Acronym, like "from" (Paging)... Those restrictions can be applied later? **These limits are by design.**

Shouldn't we limit choosing those new Entity Types (Bookmark, Acronym, chatMessage...) to be only available if the "Use beta endpoint" switch is on? **Do we have a specific user scenario for it?**

Althoug I don't know if those lists are up to date:  ExternalItem should have been supported in v1.0.

edarroudi:
hi @ZChristine , thanks for the reply. 

Regarding 

> Shouldn't we limit choosing those new Entity Types (Bookmark, Acronym, chatMessage...) to be only available if the "Use beta endpoint" switch is on? **Do we have a specific user scenario for it?**

This was meant for the PNP Search Webparts. We show the EntityTypes that work in beta only when beta mode is activiated. Otherwise the users could be confused and open issues here asking why something is not working properly. This is not a technical issue but rather prevents from extended user requests as the beta API will have some issues.
 
> Althoug I don't know if those lists are up to date: **ExternalItem should have been supported in v1.0.**

My mistake, I misread the scope table in the docs.

wobba:
@edarroudi I think we'll expose all entities by default, and people have to pick the /beta endpoint for them to light up. Makes it easier to maintain. And once it moves to v1.0 we have to re-verify anyways in case of any changes. If you feel we should check the list depending on the version, I won't block on it.

And thank you @ZChristine for replying on the pagination and support of multiple answers. We will likely look into issuing two queries as part of a batch for for this scenario then, to get both acronym and bookmark on the same user request.

edarroudi:
> @edarroudi I think we'll expose all entities by default, and people have to pick the /beta endpoint for them to light up. Makes it easier to maintain. And once it moves to v1.0 we have to re-verify anyways in case of any changes. If you feel we should check the list depending on the version, I won't block on it.
> 
> And thank you @ZChristine for replying on the pagination and support of multiple answers. We will likely look into issuing two queries as part of a batch for for this scenario then, to get both acronym and bookmark on the same user request.

I’m fine with how it’s handled right now @wobba . I would suggest approving the PR now and not splitting the query as you have to request acronyms with  specific keywords like “what is AAD”.

PR #2035, opened 2022-04-25
Feature: interface query transformation

edarroudi:
Hi guys,

I created a PR for how to implement the QueryTransformation interface as an extension. See #1947 

Added documentation now, tested as well.

Still: **An important questsion for me is whether to run the query transform before running resolveTokens or as it is like now, afterwards.
image**

As for the default query transformer: There doesn't have to be one, i just thought it could be useful to have one. I'm open to any suggestions here, maybe a possibility for users to add custom regex.

image

image

PR #2450, opened 2022-10-17
Feature: interface query transformation: Extension Interface Only

edarroudi:
as requested by @wobba i extracted the interface changes into this PR. Tested against #2035


wobba:
Are you ok to merge this @FranckyC ?

wobba:
@FranckyC makes it easier to have two PR's imo. Get the interface in, publish a new npm. Then take a dep on the new npm version in the other PR. If you have a better approach.....all for it.

FranckyC:
Doing that supposes the PR is ok to be merged which is not the case right now. What if we have to revert :), we do an other PR ? I prefer to have one to test the feature as a whole. If OK to merge, then I can publish the npm package from any branch so not a problem. I'll update the package.json file at release time when fully functional.

Issue #2461, opened 2022-10-19
Rerender on empty results when inputQuery changes


edarroudi:
**Version used**
Ex: 4.7 and forward (develop branch)

**Describe the bug**
When you serach for a text "test12" and retrieve no results, the countMessage is shown.
If you search now for "test123", the countMessage stays the same.

![image](https://user-images.githubusercontent.com/935493/196645662-eba0676f-c5f9-451f-9d8e-a5b0974752fa.png)

![image](https://user-images.githubusercontent.com/935493/196645725-2ec006b2-03df-4e28-b5c3-a5e70084c12b.png)

**To Reproduce**
see above

**Expected behavior**
Updated count message

Already fixed it and will link the PR.
![image](https://user-images.githubusercontent.com/935493/196646298-d97a2fde-29fe-4d51-a1c2-0b5a9939bbad.png)

edarroudi:
fixed in #2462 2462

PR #2462, opened 2022-10-19
Fix: Rerender on empty results when inputQuery changes

edarroudi:
Update template when inputQuery changes, regardless if data stays the same!

See #2461


wobba:
Thank you

PR #2480, opened 2022-10-24
Update mgt package references

wobba:


wobba:
@FranckyC can you test this one. Seems to work fine on my end.

Issue #2278, opened 2022-08-10
Search results: Paging doesn't work when using a page query string parameter


ileana2019:
**Version used**
Ex: 4.7.0

**Describe the bug**
On a SharePoint home page, I have a link to the search results page. The link contains a query parameter: ex: https://tata.sharepoint.com/sites/coco/SitePages/Search%20Results.aspx?Customer=blabla. On the search results page I have the search results webpart, that is configured to show paging, and to filter the results based on the Customer query string parameter:
![Search results properties](https://user-images.githubusercontent.com/55545989/183929538-ded7da2d-42ec-4955-9cec-87f5a0518fc9.PNG)

When clicking on page 2, at the bottom of the search results, nothing happens.

**To Reproduce**
1. Create a page that contains the search results webpart
2. Add a query string parameter to the page url (ex: Customer=blabla)
3. Configure the search results webpart to filter the results based on the query string parameter value (see screenshot above)
4. Save the webpart configuration and click on page 2 of the search results
5. You still see page 1 of the search results

**Expected behavior**
Display page 2 of the search results.

**Desktop (please complete the following information):**
 - Browser: Chrome

FranckyC:
@ileana2019 your configuration seems wrong. To use the query string parameter, you must use the "Use input query text" option in the connection page of the target Search Results WP like this:

![image](https://user-images.githubusercontent.com/5659122/184045507-47d146fa-7080-4dbe-9063-f9687474fb91.png)

The option "Connect to a search results Web Part" is only relevant when you connect two Search Results WPs together via item selection.

ileana2019:
HI @FranckyC ,
On the search results page, I also have a search box webpart, which I connect to the search results webpart with this settings:
![Search results](https://user-images.githubusercontent.com/55545989/184117044-4bc6b669-4a43-40ae-96fd-7138a9c7df88.PNG)
On that page I also have the filters and verticals webparts; so the entire package. The configuration that I did on the webpart was the only configuration that worked to have everything (query string parameter, search box, filters, verticals) connected to the search results webpart. 
Still, the issue remains that the query string parameter and paging do not work together.

FranckyC:
[...]

PR #2503, opened 2022-11-01
Pagination deeplink

ferrarirosso:
Fixes value mismatch when using deeplink in conjunction with filter
See https://github.com/microsoft-search/pnp-modern-search/issues/2278


ferrarirosso:
I have the bad habit to use the format document functionality of VS Code
I did only add a call to this.getFiltersDeepLink() in the componentDidUpdate method.

wobba:
Can you update the yml file.. as it should not be part of your PR.

wobba:
Thank you :)

search-parts/src/webparts/searchFilters/components/SearchFiltersContainer.tsx
```diff
@@ -134,13 +134,15 @@ export default class SearchFiltersContainer extends React.Component<ISearchFilte
...
-        this.setState({
-          submittedFilters: submittedFilters
-        });
+      const submittedFilters = this.getSelectedFiltersFromUIFilters(this.state.currentUiFilters);
+
+      this.setState({
+        submittedFilters: submittedFilters
+      });
+
+      this.getFiltersDeepLink();
```
edarroudi, line 145: this will result in count = 0 even if there is a count

wobba: @edarroudi can you propose a fix?

edarroudi: added a PR in #2548 . Moving it up lets the rest of the code run an update on the count.

Issue #2534, opened 2022-11-10
[develop 4.8 branch] Filter count bug due to PR2503


edarroudi:
**Version used**
develop 4.8 branch

**Describe the bug**
Due to #2503 a bug got in.
Count is set to zero even when there is a count.

**To Reproduce**
1. add a filter - combobox, single, withCounts
2. select an option showing any count
3. The selected option will show a count of 0

**Expected behavior**
The selected option should show the correct count

**Screenshots**
![image](https://user-images.githubusercontent.com/935493/201096943-9b2afc78-a8b1-498d-bd11-bc30211f31c9.png)
![image](https://user-images.githubusercontent.com/935493/201097249-8429ae07-765d-4a31-b8df-3a28bc710b78.png)

**Additional context**
It should work when you move  ` this.getFiltersDeepLink();` up to before `this.getFiltersToDisplay`. I mean it works then but I don't have time to test / fix it today @ferrarirosso, @wobba ...

edarroudi:
I fixed it in my PR #1845. But this is a bigger PR - if needed I create an extra PR for this BUG.

PR #2548, opened 2022-11-16
[Fix] Filter count bug due to PR2503

edarroudi:
as requested @wobba
Fixed count due to deep links calculated last.
See #2534


wobba:
Thank you :)

ferrarirosso:
Thank you !

PR #2593, opened 2022-11-25
Small documentation corrections

edarroudi:
I checked the md files and there were some small mistakes.
Also the nav link to custom data sources was wrong


wobba:
Thank you

Issue #2643, opened 2022-12-12
Search box not clearing when search suggestions enabled


Fringie:
**Version used**
4.8
**Describe the bug**
When you empty/clear the search box, the search results webpart does not update with the default value if you have query suggestions enabled, if you disable query suggestions the functionality works. I think query suggestions disables the search box if there is no input which is causing the bug.  Video example - https://www.youtube.com/watch?v=rBEOMPZMmOY

For example:

My webparts are configured to display everything via the * wildcard, so if there is nothing in the search box, everything should be displayed. But if i turn on query suggestions, it will stop working:

1. I load up the page and all the search results are displayed due to the * wildcard
2. I search for "divorce" because I'm looking for documents that use that keyword
3. I delete "divorce" because I want to reset my search
4. I expect the default search results to appear, but nothing changes. It appears that the onchange event isn't running

**To Reproduce**
(one time setup)
1. Create a page
2. Add the search box webpart to the page
3. Add the search results webpart to the page
4. Configure the search results webpart with the following settings:
Data source: SharePoint Search
Use input query text: On
Toggle the "Dynamic value" to 
- Connect to source: PnP - Search Box
- PnP - Search Box's properties: Search query
- Default value: *
5. Configure the search box webpart with the following settings:
Send the query to a new page: on
Reset query on clear: on
Enable query suggestions: on
Enable custom search providers: toggle on SharePoint Static search suggestions
6. Publish the page
(debug steps)
7. Search for anything i.e. the name of a document
8. delete the search keywords
9. At this point, the search should return to the default query because there is nothing in the search box, but it does not

**Expected behavior**
I expect that the default search results will be displayed when the search box is empty. 
 
**Screenshots**

**Desktop (please complete the following information):**
Windows 11. Chrome

**Additional context**
This issue only happens when you try to use custom search suggestions, if you turn that feature off the functionality works as expected. 
Video example - https://www.youtube.com/watch?v=rBEOMPZMmOY

Fringie:
I think I may have managed to resolve the issue. I'm still testing it but it seems to be working. The problem is the _onSearch event, I updated the if statement to the below:
![image](https://user-images.githubusercontent.com/7203124/207336982-9f76da57-23c7-4be2-9a9c-080307adda44.png)

wobba:
@Fringie here's one guide https://opensource.com/article/19/7/create-pull-request-github, and you should be able to find many on the interwebs.

Fringie:
[...]

PR #2652, opened 2022-12-19
Fix for #2643

Fringie:
There is an issue where if search box suggestions is enabled, the search box does not reload the default query if you:

Search for something
Delete what you have searched for
You cannot reload the default data; with OOTB SharePoint or with search suggestions disabled, the default data is reloaded when you have an empty search box
This change resolves the problem by allowing users to reload the default search results when the search box is empty. This is a fix for https://github.com/microsoft-search/pnp-modern-search/issues/2643, and it has caused no issues in my testing.


Fringie:
I signed the commit.. odd
image

edarroudi:
You have to create a PR for merging with the develop branch and not the main branch.

PR #2668, opened 2022-12-30
2643: Bug fix to allow the search results WebPart to display default search results when the search box is empty and search suggestions is enabled

Fringie:
There is an issue in the V4.8 Search Box WebPart when you enable search box suggestions; the search box does not reload the default query if you:

**Repro steps:**
1. Ensure Search suggestions is enabled on the Search Box WebPart
2. Add a search term into the search box then press enter
3. Delete your search term from the search box then press enter

**Expected results**
1. You would expect the default data to load as this is what happens with OOTB SharePoint
2. With search suggestions disabled, the functionality works as expected: the default data is loaded when you have an empty search box

This change resolves the problem by allowing users to reload the default search results when the search box is empty. This is a fix for https://github.com/microsoft-search/pnp-modern-search/issues/2643, and it has caused no issues in my testing.

### DIFF
--- a/search-parts/src/webparts/searchBox/components/SearchBoxContainer.tsx
+++ b/search-parts/src/webparts/searchBox/components/SearchBoxContainer.tsx
@@ -87,43 +87,40 @@ export default class SearchBoxContainer extends React.Component<ISearchBoxContai
      */
     public async _onSearch(queryText: string, isReset: boolean = false) {
 
-        // Don't send empty value
-        if (queryText || isReset) {
+        this.setState({
+            searchInputValue: queryText,
+            showClearButton: !isReset
+        });
 
-            this.setState({
-                searchInputValue: queryText,
-                showClearButton: !isReset
-            });
+        if (this.props.searchInNewPage && !isReset && this.props.pageUrl) {
 
-            if (this.props.searchInNewPage && !isReset && this.props.pageUrl) {
+            this.props.tokenService.setTokenValue(BuiltinTokenNames.inputQueryText, queryText);
+            queryText = await this.props.tokenService.resolveTokens(this.props.inputTemplate);
+            const urlEncodedQueryText = encodeURIComponent(queryText);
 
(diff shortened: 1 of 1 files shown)


Fringie:
@edarroudi I managed to create the Pull Request.

wobba:
One issue with this change is that you will execute a query always (I believe). 

You need to add logic to check if there is a default query defined, and for that case allow a query when the search box is empty.

(And OOB search will not query/search when the search box is empty, so your statement seems incorrect)

Fringie:
> you will execute a query always (I believe).

Correct.
 
> You need to add logic to check if there is a default query defined, and for that case allow a query when the search box is empty.

Interesting, I get where you're coming from, but I don't feel that's the right way to handle it. The search box submission event should always run IMO, even if you haven't set a default search query - the user pressing the Enter key means they want an action to occur. If you haven't set a default query, then you are telling the system you want the system to display nothing when the user doesn't set a term to search for IMO. 

Also, it can be unintuitive for users if, on one search experience, an empty search box returns the default search results when you submit (i.e. with query suggestions disabled), but on another search experience (i.e. query suggestions enabled), the search results don't update. 

Interested in your thoughts. 
 
> (And OOB search will not query/search when the search box is empty, so your statement seems incorrect)

I think I may have misremembered that detail so thank you for the correction. Either way, the behaviour should be similar for search box suggestions enabled and search box suggestions disabled IMO.

wobba:
My point is:
* If there is no default query set, we should NOT execute an API call for an empty query. This is waste of bandwidth and API resources on the Microsoft side
* If you add logic to check if the query (search box) is empty and there is a default query, then allow the default query to re-run.

I understand it's more logic/code, but executing an extra API call when not needed is wrong imo.

Fringie:
> Will you competent his one? If not I'll close it.

Hi sorry, hard to justify the effort on this - happy for this to be closed.

Issue #2644, opened 2022-12-15
Pagination not working on SharePoint page displayed via MS Teams personal app


PieterVandendriessche:
**Version used**
4.8

**Describe the bug**

The pagination in the search results web part is not registering my click when clicking directly on the <a> tag. When clicking left, or right my click is registered and the next result page will appear. 

This behaviour can be reproduced via an MS Teams App that has the SharePoint page as content. When using the component in the browser directly (no MS Teams) the behaviour is as expected. 

When assigning the CSS property: `pointer-events:none` to the **a** tag, the component works as expected. This indicates that the event bubbeling from the a tag to the li tag isn't working properly. ([see github repo of react-js-pagination](https://github.com/wwwaiser/react-js-pagination/blob/master/src/components/Page.js))

**To Reproduce**
Detailed steps to reproduce the behavior:
1. Open Search Page in MS Teams personal app
2. Use the pagination component by clicking on the numbers

**Expected behavior**
Clicking on the a tag should work in MS Teams as in the browser.

**Screenshots**
![image](https://user-images.githubusercontent.com/25928373/207875299-62b55a94-b3c4-405a-806b-45a42d1f924d.png)
See via with MS Teams page + desktop test:
https://user-images.githubusercontent.com/25928373/207876049-3a2a2ffa-7f77-44cb-b347-d6f8335d9042.mp4

**Desktop (please complete the following information):**
 - Ms Teams custom app

**Additional context**

Macro21:
I have the same problem... Any update?

wobba:
Fixed with the merged PR for the next release.

PR #2669, opened 2023-01-03
2644: Bug fix to allow the search results WebPart use the pagination buttons in MS Teams APP and MS Teams Web

Macro21:
Currently the "PnP Search Results" component does not work correctly in MS Teams Desktop and MS Teams Web because when clicking on the pagination buttons, they do not work. This small change in the css resolves that error.

This commit resolves error [#2644](https://github.com/microsoft-search/pnp-modern-search/issues/2644)


wobba:
Thank you :)

PR #2676, opened 2023-01-04
BUG: Fixed paging error in "MS Teams Mobile".

Macro21:
As there were many pages with 3 or 4 digit numbers in MS Teams Mobile, the numbers at the end were not displayed, so they have been moved down to the next row.

For example: << < 281 282 282 283 284 .... 290 ... 310 > >> in Teams Mobile was cut off and you could not click on 310 for example.

With this solution in MS Teams Mobile it will be: 

<< < 281 282 283 284 ...
290 ...301 ..
310 > >>

BUG:
image

Solution: 

image

### DIFF
--- a/search-parts/src/components/PaginationComponent.module.scss
+++ b/search-parts/src/components/PaginationComponent.module.scss
@@ -18,7 +18,10 @@
 
             ul {
                 display: flex;
+                flex-wrap: wrap;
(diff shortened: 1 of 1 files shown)


wobba:
Thank you, and you might want to change the setting to display max 5 pages in the range - which is a better user experience.

Issue #2289, opened 2022-08-16
Search Results Web Part : "Select all" checkbox in the "Details List" layout doesn't trigger connected web parts update


Titef:
**Version used**
v4.6.1

**Describe the bug**
When two Search Results Web Parts are connected and you use the "Details List" layout for the data provider Web Part, using the "Select All" checkbox in the headers will not trigger the Consumer Web Part to update.

**To Reproduce**
1. Add a Search Result Web Part WebPart1 with, for example, this Search Query : {searchTerms} IsDocument=true FileExtension="pptx"
2. Set WebPart1 layout to "Details List"
3. Add another Search Result Web Part WebPart2 with, for example, this Search Query : {searchTerms} contentclass="STS_Site"
4. Set WebPart2 to use data from WebPart1, connecting for example "SiteId" from Provider to "SiteId" from Consumer
5. Select a result within WebPart1 -> WebPart 2 is updated
6. Click on "Select All" checkbox in the WebPart1 header to select all results -> WebPart 2 is not updated

**Expected behavior**
The "Select all" checkbox of the "Details list" layout should trigger an update for connected Web Parts

**Screenshots**
![image](https://user-images.githubusercontent.com/34166028/184816232-8f79ef80-56f1-4247-a907-b8df6132d7da.png)
![image](https://user-images.githubusercontent.com/34166028/184815970-3c9df1be-40d3-42a8-9332-9449c7f5a97b.png)
![image](https://user-images.githubusercontent.com/34166028/184816389-32e1769b-c303-4887-b813-aa91bb41608a.png)

**Desktop:**
 - Browsers: all

wobba:
Please re-test with the latest version to verify it's still an issue.

Titef:
Hello,
Same issue in v4.7.0.
Thanks

wobba:
Fixed with PR from @patrikhellgren .

PR #2678, opened 2023-01-05
Fix for Select All button in DetailsList

patrikhellgren:
There is an issue with how the SelectionZone in the search results container is implemented together with the DetailsList internal selection handling. Clicking Select All in the DetailsList selects all items in the list but does not update selectedKeys in the templateContext.

This is also the issue behind #2289 so that should also be fixed after this.

This fix adds the ```data-selection-all-toggle=true``` property to the div element containing the Select All button which makes the search results container SelectionZone also trigger on the button click.


wobba:
Thank you.

PR #2679, opened 2023-01-05
Correct data-selection-all-toggle documentation

patrikhellgren:
Corrects the documentation for the ```data-selection-all-toggle``` property. The Fluent UI documentation is also incorrect on this, see [code](https://github.com/microsoft/fluentui/blob/4afa85e9437c0ccdf33cdccce48f5b3aa8da138a/packages/react/src/utilities/selection/SelectionZone.tsx#L40).


wobba:
Thank you

PR #2737, opened 2023-02-03
New scenario Use query string from url for dynamic results

davemehr:
This scenario describes how to use query string as value in the URL from the current page.


wobba:
Awesome!!

PR #2841, opened 2023-03-10
Changes to swedish translations

patrikhellgren:
A few changes and corrections for the swedish translations.


search-parts/src/webparts/searchResults/loc/sv-SE.js
```diff
@@ -17,11 +17,11 @@ define([], function() {
         PagingOptionsGroupName: "Inställningar för sidindelning",
         ItemsCountPerPageFieldName: "Antal objekt per sida",
         PagingRangeFieldName: "Antal sidor som ska visas",
-        ShowPagingFieldName: "Antal sidor som skall visas",
+        ShowPagingFieldName: "Antal sidor som ska visas",
```
wobba, line 20: Would `Antal sidor att visa` be better?

patrikhellgren: Good that you asked. After double checking the english localization file I realized that the original swedish translation for ShowPagingFieldName was totally wrong so I was doing a small correction for something that actually needed to be redone altogether. I have now changed both ShowPagingFieldName and PagingRangeFieldName to something that has more meaning in the context and with inspirations from the norwegian translations which looked much better.

wobba:
Thank you :)

Issue #2842, opened 2023-03-10
Security vulnerabilities reported by Whitesource


erkanch:
Whitesource reported the following security vulnerabilities: 
https://1drv.ms/x/s!An0A9oJjn3HHgc9seQL41IDZO4jrBw?e=JF2CgN

wobba:
PR merged

PR #2843, opened 2023-03-10
Address vulnerabilities identified by Whitesource

erkanch:
Fix for #2842 
Addresses the the following security vulnerabilities reported by Whitesource:
https://1drv.ms/x/s!An0A9oJjn3HHgc9seQL41IDZO4jrBw?e=JF2CgN


wobba:
Thank you, and want to call out that `set-value` is the only lib in the list being part of the built bundle. The rest are part of SPFx build chain only.

PR #2881, opened 2023-03-29
Update nb-no.js

kjetilodegaarden:
The norwegian translation of default search option is wrong. Changed it to "Default" instead of "Misligholde" which makes no sense with regards to sorting.


wobba:
"Default" is English. I would rather have "Standard" which aligns with the Danish and Swedish translations.

kjetilodegaarden:
Agreed. Norwegian customer wanted default so did not think about it. Standard is better.

Issue #2933, opened 2023-04-27
People Layout Is Broken If The Top Level Domain is Longer Than 5 Characters


Lotsamocha01:
**Version used**
Ex: 4.8

**Describe the bug**
When using the People layout, any email address with a top level domain longer than 5 characters is clipped, preview images are not shown and the name displays as "PersonImmersive.aspx".

**To Reproduce**
Detailed steps to reproduce the behavior:
1. Have an e-mail address with a top level domain longer than 5 characters. Ex. "xxxx@kbc.global"
2. Create a search results web part that uses the People layout
3. The name shows as "PersonImmersive.aspx", no profile image, and the e-mail address is clipped at the end

**Expected behavior**
The user name should display with a profile image and the full e-mail address

**Screenshots**
![image](https://user-images.githubusercontent.com/114703957/234901193-ca178692-f52e-4023-8d6a-af94532df791.png)

**Desktop (please complete the following information):**
 - Browser is Chrome, Edge, and Firefox

**Additional context**
I suspect the issue is due to a Regex expression in the TemplateService.ts file that limits the TLD to a max of 5 characters

[https://github.com/microsoft-search/pnp-modern-search/blob/main/search-parts/src/services/templateService/TemplateService.ts#L754](url)

`const matches = expr.match(/([a-zA-Z0-9_\-\.]+)@([a-zA-Z0-9_\-\.]+)\.([a-zA-Z]{2,5})/gi);`

wobba:
Feel free to PR a fix.

wobba:
Fixed with #2938

PR #2938, opened 2023-04-27
Fix tld length

Lotsamocha01:
Updated the Regex expression to allow for 2-63 octets. As I understand it, 63 octets is the max per label according to RFC 1034 but I'm not very familiar with Regex so please let me know if the change is feasible.

#2933


wobba:
Thank you.

Issue #911, opened 2021-04-02
Documentation clarification question regarding {{unique}}


bryllyg:
Hi, I'm having difficulties understanding what and how the {{unique}} handlebar helper works. I have a query that returns a list of names, some of which are listed multiple times, and I'd like to be able to eject the duplicates. I though {{unique}} sounded promising, but I've not been able to get it to work. Is this what it actually does? Or am I not understanding it's use. Can it not be used to adjust items retrieved during an {{#each}}?

I'm attempting to iterate through;
{#each items as |item|}}
   {{nameOWSTEXT}}
{/each}

My results looks like:
Jane Doe
Jane Doe
Sara Smith
Jane Doe

I'd love to nip that down to just the two unique entries. I've tried about a zillion different ways to insert {{#each (unique)}} into the equation with no success. If it is doable, I could really use an example. Is there anywhere I can be looking for more examples to pour over? It would be very helpful.

Thanks much!

wobba:
Don't have a sample at hand but if you only output one field you can try to use `pluck` https://github.com/helpers/handlebars-helpers#pluck. Another approach is `sortby`, and in the loop compare with the previous item and only render if it is different.

The better approach was if we exposed `collapsespec` as part of web part settings which would only return unique values, and you would not have to handle this as part of the UX loop.

bryllyg:
Thanks, I was looking at pluck and sortby too, yah! I'm getting the hang of it. Many thanks for the pointers and clarification.  I may add a feature request for collapsespec for future revs, that would be handy. Off to go whip it into shape!

bryllyg:
I did some more testing, and it looks like my problem may be that pluck and unique are not getting loaded. I get no results from my query, and if I F12 and look at the Console log, I see "Uncaught (in promise) ... Missing helper: "pluck""

![image](https://user-images.githubusercontent.com/2616432/113602069-00ee2580-95f7-11eb-8baa-3e2b768dc57a.png)

I've been able to use other helpers such as 'has' and 'lowercase', so I've been assuming I don't have to install anything extra to use all of the helpers listed here: https://github.com/helpers/handlebars-helpers. Am I incorrect?

wobba:
I'll recheck if they are included or not. We are using the latest version and I'll look at this and report back.

bryllyg:
[...]

PR #3043, opened 2023-05-30
Added collapse specification property

svenbru:
Added the ability to configure the collapse specification property in the SharePoint data source, like mentioned in older issues like #2135 and #911 

CollapseSpecification


wobba:
Care to add it for the Graph provider as well? It's on the beta endpoint named `collapseProperties`.

https://learn.microsoft.com/en-us/graph/api/resources/searchrequest?view=graph-rest-beta

svenbru:
@wobba Sure! I made the changes for the Graph provider.

search-parts/src/dataSources/MicrosoftSearchDataSource.ts
```diff
@@ -787,6 +839,10 @@ export class MicrosoftSearchDataSource extends BaseDataSource<IMicrosoftSearchDa
             searchRequest.trimDuplicates = this.properties.trimDuplicates;
         }
 
+        if (this.properties.collapseProperties.length > 0) {
```
wobba, line 842: This should go under the `this.properties.useBetaEndpoint` for now.

svenbru: Makes sense, shouldn't the trimDuplicates property also be set only when useBetaEndpoint is enabled as this property is only exposed in the beta endpoint? I've included this in my latest commit.

wobba: It should, and that one will hopefully/maybe never make it's way to v1.0. Thank you 🙂

wobba:
There are some merge conflicts on loc files to resolve.

svenbru:
Probably had to do with some auto formatting in VS Code, The conflicts should be corrected now.

PR #3061, opened 2023-06-09
Fix fast-xml-parser vulnerability

erkanch:
Addresses the following vulnerability: https://github.com/NaturalIntelligence/fast-xml-parser/security/advisories/GHSA-6w63-h3fj-q4vw


erkanch:
Just realized someone has already created a PR for this

PR #3067, opened 2023-06-11
Added more date/time tokens

wobba:
#3047


EMaarten:
Hi @wobba ,

I've noticed that the tokens CurrentHourUTC, CurrentMinuteUTC and CurrentSecondUTC can return 1 digit values.
This breaks the search query as leading zeros should be included otherwise no search results were found.

For the tokens CurrentMonthUTC and CurrentDateUTC there's no impact on the search query when a 1 digit value is returned by the token.

Kind regards

PR #3170, opened 2023-08-02
Fix for web part showing empty when URL has two letter code

RobertSStorm:
When the URL of a web page contains a doc library with a two letter code that is not an ISO code (sites/site-name/SitePages/xx/Page.aspx), the search web parts on this page would show empty.


RobertSStorm:
Replaced by #3172

Issue #3157, opened 2023-07-24
Filters read from url are not set in the filter web part


djsladi:
**Version used**
4.9.1

**Describe the bug**
When saving URL with filters, the filter panel does not have the filters (coming from URL) selected

**To Reproduce**

1. Set search results and filters web part to a page
2. Select filter(s)
a. ![image](https://github.com/microsoft-search/pnp-modern-search/assets/39775012/83c1ffba-cd6b-4e30-a4c6-b405b582ab19)
3. Copy URL with filters added
a) pagename.aspx?**f=%5B%7B"filterName"%3A"RefinableString140"%2C"values"%3A%5B%7B"name"%3A"Lorem%20Ipsum%20documents"%2C"value"%3A"ǂǂ4c6f72656d20497073756d20646f63756d656e7473"%2C"operator"%3A0%2C"disabled"%3Afalse%7D%5D%2C"operator"%3A"and"%7D%5D#**
4. Paste URL back to the address bar
5. Search results are filtered by variables from URL, the filter panel shows only selected filters but they are not selected, thus making the clearing filters impossible (without selecting the filter again)
a) 
![image](https://github.com/microsoft-search/pnp-modern-search/assets/39775012/9124e63a-4687-442c-9d22-ed13ca623ac0)

**Expected behavior**
Filters should be applied already, making clearing them more reasonable:
![image](https://github.com/microsoft-search/pnp-modern-search/assets/39775012/5dca769c-dec9-45a3-847b-8a918e253bda)

**Desktop (please complete the following information):**
 - Browser Microsoft Edge, Version 115.0.1901.183 (Official build) (64-bit)

mbice:
Seeing the same issue here. We use URL-based filtering frequently for things like clickable metadata tags as well as for navigation (many of our customers have links with filters already on the URL and users also sometimes bookmark refined search pages to save clicks).

kasperbolarsen:
I'll look into this over the weekend unless somebody else is looking into it?

kasperbolarsen:
It is correct that the filters are not selected, neither for single nor multiselect values. However, the available filter values are reduces to the filters specificed in the URL. As a workaround the users can select those filter values, and then unselect them again to return to the inital state. Another option could be to add a Button web part near the search box with the label "Reset search" and the URL in the Button web part will reload the page without search parameters.
I have marked this issue as an Enhancement rather than a bug. Thanks for pointing out the issue :-)

mbice:
[...]

Issue #2624, opened 2022-12-05
Filters rendering


LRPowys:
**Version used**
Version 4.8

**Describe the bug**
Upon selection of values, associated with the first two filters, it hides the rest of the filters. The filters reappear if the browser page is refreshed. The filters are 5 in total and correspond to columns with the same names as created in the document libraries.

**To Reproduce**
Detailed steps to reproduce the behavior:
1. Go to the 'Filters' webpart of the PNP Module
2. Click on the first checkbox to select a value from the first filter. Then click on the next checkbox down to select a value from the second filter. The rest of the filters, visible up to that moment then disappear and it is not possible to further specify the filtering. There are definitively tags on files that are in Level3 filter and above.
3. Go to browser 'Refresh' at the top or press CTRL+R.
4. The missing filter(s) reappear and a value within it can now be selected.

**Expected behavior**
The filters should not hide if there are values associated with files within that particular filter. Constant refreshing of the browser page to prompt the filters to re-appear is not acceptable from user perspective. It used to be the filters were working properly and the fault seems to have started occuring after the upload of some files on the site. The upload may just be a coincidence and may not have anything to do with the issue.

**Screenshots**

Screenshot 1
![image](https://user-images.githubusercontent.com/119879292/205687869-dc3537e3-0d89-4780-abc2-aa30d4745132.png)

Screenshot 2
![image](https://user-images.githubusercontent.com/119879292/205703854-f8e3ea29-73e6-485d-87ef-9bba075b31e5.png)

Screenshot 3
![image](https://user-images.githubusercontent.com/119879292/205688561-9ff9690b-d3f4-409a-a4cb-878a2e3a9d32.png)

Screenshot 4
![image](https://user-images.githubusercontent.com/119879292/205689097-4bbb1ae3-79ae-4151-867d-ff41c9796ccf.png)

Screenshot 5
![image](https://user-images.githubusercontent.com/119879292/205689332-e41d7621-2c79-4630-b86d-42fbc29e2843.png)

**Desktop (please complete the following information):**
 - Browser: Chrome, Edge

**Additional context**
[...]

PR #3180, opened 2023-08-07
#3157: Bugfix Filters read from url are not set in the filter web part

salimbharuchi:
This bug happened due to the regression of #2624


wobba:
Thank you :)

PR #3305, opened 2023-10-25
Fix searchTerms after clearing search box

### LINKED ISSUE #3197: Clearing SearchBox doesn't work properly when using a query with {searchTerms}
**Version used**
4.9.2

**Describe the bug**
Clearing the SearchBox doesn't actually clear the query text when using {searchTerms} in the Query Template with SharePoint Search. 

**To Reproduce**
1. Add a Search Box WebPart - leave at default configuration
2. Add a Search Results WebPart - with only the following changes
2.1. Use SharePoint Search as the data source
2.2. Change the query template to **{searchTerms} PromotedState:0**
2.2.1. This will ensure you should get results even with an empty queryText
2.4. Connect the Search Box WebPart via Available connections with the following configuration
image

3. Publish the page
4. You should now get a results of all the content pages in your tenant
5. Perform e.g. a single word search in the Search Box that should give you less results but some
6. Note the Results count part of the Search Results WebPart shows less results and the query text
7. Clear the Search Box with the X button on the Search Box
8. The Results count doesn't show the query text anymore, but you are still getting the same results as before clearing the Search Box

**Expected behavior**
You should get the results as in step 4

**Ways to circumvent this**
If you change the query template to **{searchBoxQuery} PromotedState:0** this bug won't reproduce.

**Additional Context**
This seems to be a regression that came with #3028.  When you initially perform a search, TokenService is setting {searchTerms} with the text set in the Search Box. After clearing the Search Box the query still uses the old value in the querytemplate. My  first approach here would be to completely comment out the following lines:

https://github.com/microsoft-search/pnp-modern-search/blob/284e2bd8d1d8a1066d156e7224f5f10c106ca89c/search-parts/src/webparts/searchResults/SearchResultsWebPart.ts#L2246-L2249

I'm not sure if this is applicable when using another data source other than SharePoint though. SharePoint Search service will properly set the {searchTerms} under the hood if provided with a querytext. I quite often use {searchBoxQuery} instead of {searchTerms} as I noticed the same issue that #3028 was supposed to fix.

jhholm:
Fixes #3197

This fix changes the logic to set the original token value undefined for searchTerms if you clear your search box and thus inputQueryText is falsy.

### DIFF
--- a/search-parts/src/webparts/searchResults/SearchResultsWebPart.ts
+++ b/search-parts/src/webparts/searchResults/SearchResultsWebPart.ts
@@ -2209,7 +2209,7 @@ export default class SearchResultsWebPart extends BaseWebPart<ISearchResultsWebP
             this.tokenService.setTokenValue(BuiltinTokenNames.originalInputQueryText, inputQueryText);   
 
             //Set searchTerms value from inputQueryText, but set initial token value, undefined, if empty
-            const searchTerms = inputQueryText ? inputQueryText : undefined;
+            const searchTerms = inputQueryText ?? undefined;
             // Legacy token for SharePoint and Microsoft Search data sources
             this.tokenService.setTokenValue(BuiltinTokenNames.searchTerms, searchTerms);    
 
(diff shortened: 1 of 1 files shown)


wobba:
Thank you :)

Issue #2972, opened 2023-05-16
Performance fixes for card template when showing many results on one page


jhholm:
**Performance woes when using card template and e.g. 100 results**
Page loads and pagination changes can be quite slow, and especially on edit mode even on a rather recent PC.

**Description of use case**
Quite common use case is to use PnP Modern Search as a navigation page. E.g to show all project sites that are available. When using the handlebars card template and showing e.g. 50 to 100 results on a page, the generated HTML becomes quite massive. This makes the page load quite slow on lower end systems and mobile. This is probably because each web component (pnp-documentcard) contains a data-context attribute that has all the search result items and some additional duplicated data that are present in other attributes (e.g. theme).

**Examples:**
1. Showing 100 results, default selected properties, compact card template: 320 000 characters per web component => ~32 million characters of HTML for 100 items
2. Removing all other result items from data-context with same scenario in 1 => 26 000 characters per item  => 260 thousand characters of HTML for 100 items

**Describe the solution you would recommend**
I think it might be feasible to refactor what is provided to data-context or refactor the template somewhat. Should this actually be a discussion instead of an issue? 

**I am willing to do a pull request on this, but would actually want to get some feedback what kind of change seem most feasible.**

**Additional things to consider**
I guess it might also be worth to take a look if the images could be lazily loaded when they appear on viewport as that might also factor on the performance.

wobba:
Feel free to propose an optimization :)

jhholm:
I tried to understand the current implementation and come up with a workaround that would not create any breaking changes. I think the context object contains currently duplicate data, as theme is usually provided as another attribute, but it's still used in the template processing.

**data-context** seems to be used only by the following web components
* pnp-documentcard
* pnp-detailslist
* pnp-persona
* pnp-slider

=> Thus JSONStringify @root is only used in those places

**pnp-documentcard** and **pnp-persona** use **ITemplateService processFieldsConfiguration** to render the handlebars template
https://github.com/microsoft-search/pnp-modern-search/blob/cda9ca9fc0ccd3068f1b64ab8b68e2a5c9d4a516/search-parts/src/components/DocumentCardComponent.tsx#L113
https://github.com/microsoft-search/pnp-modern-search/blob/cda9ca9fc0ccd3068f1b64ab8b68e2a5c9d4a516/search-parts/src/components/PersonaComponent.tsx#L100

The following properties are explicitly used in both cases
[...]

PR #3306, opened 2023-10-25
Context truncation

jhholm:
Regarding #2972 

I reworked how the truncation works, that it just drops items and uses destructuring.


wobba:
Thank you :)

PR #3414, opened 2023-12-18
Include WebPartTitle for SearchBox WebPart

kachihro:
To ensure consistency for the SearchResults, Filter and Verticals - I've included a TITLE for the WebPart for SearchBox.

This is especially useful on a HOME PAGE - when needing to 'send query to new page'.


kachihro:
Have configured PGP to ensure the code commit is verified

kachihro:
This is the end result - with the ability to have a webpart title - rather than a separate TEXT block (and big space/gap)

image

kachihro:
** How do I sign/verify the 'prior' commits ??

search-parts/src/webparts/searchBox/components/SearchBoxContainer.tsx
```diff
@@ -146,6 +147,16 @@ export default class SearchBoxContainer extends React.Component<ISearchBoxContai
     }
 
     public render(): React.ReactElement<ISearchBoxContainerProps> {
+
+        let renderTitle: JSX.Element = null;
+
+        // WebPart title
+        renderTitle = <WebPartTitle
```
wobba, line 154: If a property is not filled out, will this render with no pixels taken?

kachihro: Correct.   I've checked the output/html - and no title has zero impact to the rendered content.   No additional pixels.

image

Issue #3421, opened 2023-12-22
Filters on taxonomy managed properties don't show selected value correctly


mbice:
**Version used**
4.10.1

**Describe the bug**
This is a new issue with 4.10.1. When applying a filter on a taxonomy managed property, there are a couple of issues I can see:

1. The text on the selected tag is dark and not accessible
2. If "Multi values" option is un-checked (so we are using a single select taxonomy filter), the selected filter's radio button is not checked.

**To Reproduce**

1. Add search results and filters web parts and wire them up together
2. Add a refiner that is mapped to a taxonomy field
3. Ensure "Multi value" is not checked
4. Apply a filter

**Screenshots**
![image](https://github.com/microsoft-search/pnp-modern-search/assets/985283/78760630-2913-424f-8ffc-bc891b04aa13)

**Desktop (please complete the following information):**
 - Windows, Chrome (latest)

kasperbolarsen:
The color on the selected filter value seems for follow the site theme. 
Whether the non-selected filter values disappear or becomes grayed out seems to depend on any other filter being applied

initial ( no filters applied)
![image](https://github.com/microsoft-search/pnp-modern-search/assets/20593570/17420426-6179-4f74-9b13-6e3160dc50f6)

Top filter applied
![image](https://github.com/microsoft-search/pnp-modern-search/assets/20593570/3942a1b2-be56-4840-b736-6b8d1a3233a3)

Both filters applied
![image](https://github.com/microsoft-search/pnp-modern-search/assets/20593570/71bda480-a6b7-4d39-a063-0c9cc4d10585)

Only buttom filter applied
![image](https://github.com/microsoft-search/pnp-modern-search/assets/20593570/03b09ef5-a36b-4e1b-ae7a-890390db36d6)

PR #3429, opened 2024-01-05
Fix taxonomy filters selected tag color and radio button checked state

mbice:
Fix selected filter tag text color and ensure checked state of radio button is applied. Applies to #3421.

### DIFF
--- a/search-parts/src/components/filters/FilterSearchBoxComponent.module.scss
+++ b/search-parts/src/components/filters/FilterSearchBoxComponent.module.scss
@@ -10,14 +10,14 @@
     flex-wrap: nowrap;
     user-select: none;
     background-color: "[theme: themePrimary, default: #0078d7]";
-    color: #fff;
+    color: "[theme: white, default: #ffffff]";
     padding-left: 5px;
     border-radius: 12px;
     margin-bottom: 6px;
     max-width: 100%;
 
     .tagItemText {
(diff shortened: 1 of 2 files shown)


search-parts/src/components/filters/FilterSearchBoxComponent.module.scss
```diff
@@ -17,6 +17,7 @@
     max-width: 100%;
 
     .tagItemText {
+        color: #fff;
```
wobba, line 20: Should this use the theme color for white instead? Which would then be black on high contast.

mbice: @wobba good point. I updated it now to use the theme, instead. Thanks!

mbice: @wobba sorry, one other thing. If this PR is good and gets merged in, do you think there would be any chance to get a 4.10.2 patch release?

wobba:
Thank you

search-parts/src/components/filters/FilterSearchBoxComponent.module.scss
LeonArmston: Great work @mbice - I'd love this in a 4.10.2 release - I've got a client this is affecting at the moment.

PR #3432, opened 2024-01-09
Update scenario index.md

davemehr:
The title and link to the scenario ## [Use query string from url for dynamic results](use-query-string-in-url.md) was missing.


wobba:
Thank you

PR #3455, opened 2024-01-15
add breadcrumb component for spo entities

lauribohm:
## Breadcrumb component

The component is designed to construct a breadcrumb path, ideally utilizing the `OriginalPath`, `SPSiteURL`, `SPWebUrl`, `Title`, and `FileType` properties derived from a SharePoint entity (such as a file, item, folder, document library, etc.). Component validates that neither of `OriginalPath`, `SPSiteURL`, `SPWebUrl`, or `Title` aren't `undefined` or `null` and therefore component is designed to work for SharePoint entities. 

The actual breadcrumb is constructed using the [Fluent UI breadcrumb](https://developer.microsoft.com/en-us/fluentui#/controls/web/breadcrumb).

Breadcrumb path should work fine and look good for:

- files
- pages
- lists
- list items
- sites
- document libraries
- folders

###  User instructions

- **Description**: Render a breadcrumb path of a SharePoint entity (file, item, folder, document library etc.).

breadcrumb-component

- **Usage**: 

Get started with:
```html
<pnp-breadcrumb 
    data-path="{{OriginalPath}}"
    data-site-url="{{SPSiteURL}}"
    data-web-url="{{SPWebUrl}}"
    data-entity-title="{{Title}}"
    data-entity-file-type="{{FileType}}"
/>
```
Use all properties:
```html
<pnp-breadcrumb 
    data-path="{{OriginalPath}}"
    data-site-url="{{SPSiteURL}}"
    data-web-url="{{SPWebUrl}}"
    data-entity-title="{{Title}}"
    data-entity-file-type="{{FileType}}"
    data-include-site-name="false" 
    data-include-entity-name="true"
    data-breadcrumb-items-as-links="true"
    data-max-displayed-items="3"
    data-overflow-index="0"
    data-font-size="12"
/>
```
|Parameter|Description|
|--|--|
|data-path|Used for creating the breadcrumb path. Component is designed to receive `OriginalPath` or `Path` property. Property is required for rendering the breadcrumb path. `String`|
|data-site-url|Used for creating the breadcrumb path. Component is designed to receive `SPSiteURL` property. Property is required for rendering the breadcrumb path. `String`|
|data-web-url|Used for creating the breadcrumb path. Component is designed to receive `SPWebUrl` property. Property is required for rendering the breadcrumb path. `String`|
|data-entity-title|Used for creating the breadcrumb path. Component is designed to receive `Title` property. Property is required for rendering the breadcrumb path. `String`|
|data-entity-file-type|Used for creating the breadcrumb path. Component is designed to receive `FileType` property. Property is required for rendering the breadcrumb path. `String`|
|data-include-site-name|If the site name should be included in the breadcrumb items. Optional, default value `true`. `Boolean`|
|data-include-entity-name|If the entity name should be included in the breadcrumb items. If the value is set to `false`, not only is the entity name excluded from the breadcrumb path, but also the last item in the breadcrumb path is not highlighted in bold. Optional, default value `true`. `Boolean`|
|data-breadcrumb-items-as-links|If the breadcrumb items should be clickable links to the path they represent. Optional, default value `true`. `Boolean`|
|data-max-displayed-items|The maximum number of breadcrumb items to display before coalescing. If not specified, all breadcrumbs will be rendered. Optional, default value `3`. `Int`|
|data-overflow-index| Index where overflow items will be collapsed. Optional, default value `0`. `Int`|
|data-font-size|Font size of breadcrumb items. Optional, default value `12`. `Int`|


search-parts/src/components/SpoPathBreadcrumbComponent.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+    /**
+     * The current theme settings.
+     */
+    themeVariant?: IReadonlyTheme;
+}
+
+export interface IBreadcrumbState { }
+
+const SITE_REGEX = /https:\/\/\w+\.sharepoint\.com\/sites\//;
```
wobba, line 51: What about /teams? Also some customers have custom domains and government clouds use different domains as well. It's better to use the SPWebUrl as the base when start splitting to avoid dealing with domains.

wobba:
You have to update documentation as well as part of the PR with screenshot and how to use the component.

search-parts/src/components/SpoPathBreadcrumbComponent.tsx
lauribohm: I've now included case /teams in regex check. But per my understanding using domain extensions such as *.mcas.ms, *.mcas-gov.us, or *.mcas-gov.ms does not impact this validation as they are not part of the OriginalPath property.

Also added prop for SPWebUrl and used that while creating the breadcrumb path.

lauribohm:
> You have to update documentation as well as part of the PR with screenshot and how to use the component.

Added instructions how to use the component. Let me know if even more comprehensive documentation is needed.

search-parts/src/components/SpoPathBreadcrumbComponent.tsx
wobba: Please address my URL comment as well. Don't use `sharepoint.com`in the regex, but rather pass in the `SPWebUrl` value or similar and use that to massage the URL.  The domain could e.g. be `sharepoint.us` or even `contoso.com` for custom domain customers.

lauribohm: Okey, modified the component and added comments how I think it should work.

search-parts/src/components/SpoPathBreadcrumbComponent.tsx
```diff
@@ -0,0 +1,169 @@
...
+    private getBreadcrumbItems = (spWebUrl: string, originalPath: string, includeSiteName: boolean, includeEntityName: boolean, breadcrumbItemsAsLinks: boolean): IBreadcrumbItem[] => {
+        // spWebUrl: https://contoso.sharepoint.com/sites/sitename/subsite
+        // originalPath: https://contoso.sharepoint.com/sites/sitename/subsite/Shared Documents/Document.docx
+
+        const frags = spWebUrl.split('/');
+        // frags: ["https:", "", "contoso.sharepoint.com", "sites", "sitename", "subsite"]
+        
+        const basePath = frags.slice(0, 4).join('/');
+        // basePath: https://contoso.sharepoint.com/sites
+
```
wobba, line 123: @lauribohm Do you handle the root site as well with this code?

lauribohm: Yes... Made a new commit and added root site handling.

wobba:
One more question before I merge. Should we add this component to one of the default templates you think?

lauribohm:
> One more question before I merge. Should we add this component to one of the default templates you think?

Yes, it would make sense to include this component in the details list template, and possibly in the list template as well. 

Additionally, I made an extra commit to handle entities with `DispForm.aspx?ID=xx` in their path, such as images and list items. As a result, the breadcrumb path will display the title + possibly extension/file type of those entities as the last part, instead of `DispForm.aspx?ID=xx`.

wobba:
Thank you.. if you acknowledge it looks good for:
* files
* pages
* list items
* sites
* folders
then we are good (don't have time to build myself right now)

lauribohm:
> Thank you.. if you acknowledge it looks good for:
> 
> * files
> * pages
> * list items
> * sites
> * folders
>   then we are good (don't have time to build myself right now)

Aah yes, noticed that lists and document libraries don't have webUrl which was causing breadcrumb to not render for those. Removed webUrl null check from props validations and it is now using siteUrl to help create the breadcrumb if webUrl is null. Seems to work.

wobba:
Thank you.

PR #3518, opened 2024-02-08
Patch 8 - Fixing "Issues" List Link

HiltonGiesenow:
It's currently pointing to a 404. Presumably it should point to the actual issues list for this project?


wobba:
Thank you :)

PR #3532, opened 2024-02-14
Fixed some typos and texts in German localization files

jblueml:
Fixed some typos and reworked texts in German localization files.


jblueml:
I was just wondering, why the SonarCloud-check failed as I neither duplicated something nor added something new. I have only changed some texts.

wobba:
Thank you :)

PR #3545, opened 2024-02-20
Added missing localization-strings for German language

jblueml:
I found out that a few localization strings for the German language were missing (e.g. in sort settings).
This also resulted in the option "Show persona card on hover (LPC)" not being displayed in the property pane.


wobba:
Thank you

Issue #3505, opened 2024-02-07
Search Results and Search Filters webparts doesn't load using mobile device


sherpas-par:
**Version used**
4.10.2.0

**Describe the bug**
When using mobile device, in my case Apple iPhone, Search results and Search Filters webparts doesn't load, only showing error: undefined is not an object (evaluating 'e.isInternal').

Used these web browsers, Apple Safari and Microsoft Edge. I've also tried using Microsoft SharePoint App for iOS.

**To Reproduce**
Detailed steps to reproduce the behavior:
1. Go to SharePoint Online page with configured Modern Search webparts, Search Results and Search Filters using mobile device
2. See error

**Expected behavior**
Search Result and Search Filters webpart should load correctly and display content.

**Screenshots**
Error message when using 4.10.x.0
![image](https://github.com/microsoft-search/pnp-modern-search/assets/57943258/5264ca4a-3153-4b96-9908-579517ec3d5c)

**Desktop (please complete the following information):**
Apple iOS 17.3
 - Browser Safari and Edge (both are running Apple WebKit)

**Additional context**
Tried Modern Search 4.9.3.0 and I **cannot** reproduce the problem with that version.
![image](https://github.com/microsoft-search/pnp-modern-search/assets/57943258/e605958f-e9da-4a61-8783-ad14540e2099)

I've also tried Modern Search 4.10.1.0 and the **same issue** occurs as with 4.10.2.0

kasperbolarsen:
I tried to recreate the issue on my Samsung s22, Android 14, and I had no issues at all.

sherpas-par:
I have tried now with two different phones:

- Samsung S22, Android 14, Google Chrome - **Working**
- OnePlus 9 Pro, Android 13, Google Chrome - **Not Working**

wobba:
I can repro the issue on my iPhone, but not sure how to debug and fix it... as the error is coming from sppages and not the web parts themselves.

wobba:
Seems to work if I turn of use of GraphToolkit, so related to that somehow.

DavidD410:
> Seems to work if I turn of use of GraphToolkit, so related to that somehow and not sure I want to dig into this at all :)

Can you clarify this, please ?

patrikhellgren:
I can reproduce this using Edge on Windows by clearing the cache and reloading the page.
And also by instead adding sw=bypass to the querystring.

wobba:
[...]

PR #3556, opened 2024-02-22
Upgrade to mgt4

patrikhellgren:
Upgrades MGT to version 4. This resolves the issue reported in #3505.
This also introduces two new settings for MGT disambiguation to make it possible to run different versions of MGT on the same page without the use of mgt-spfx that has been deprecated in MGT 4. One setting is for the general _Use Microsoft Graph Toolkit_ setting and the other is for the _Show Persona hover card_ setting since there seems to be no way to get to the main webpart properties from the layouts.

Unfortunately there has been some changes to the mgt components property naming and styling that will require changes in already existing custom layouts/columns. E.g. for mgt-person the _oneLine_ property has changed casing to _oneline_ and the font styling has also changed. The default Person layout should work when using no disambiguation since it uses mgt-person.


search-parts/src/layouts/results/people/PeopleLayout.ts
```diff
@@ -170,6 +175,13 @@ export class PeopleLayout extends BaseLayout<IPeopleLayoutProperties> {
                 offText: strings.General.OffTextLabel,
                 checked: this.properties.showPersonaCard,
             }),
+            PropertyPaneTextField('layoutProperties.microsoftGraphToolkitDisambiguation', {
```
wobba, line 178: @patrikhellgren do we want this as user set string, or can we set it via code? E.g. use webpart id or something.

patrikhellgren: @wobba While researching a response to your comment I found that you can't disambiguate mgt components between instances of the same web parts. So we have to have the same disambiguation string for the whole package. This meens we can hard-code a string to be used for the PnP Modern Search solution. This should not be a problem as long as one doesn't manage to put webparts from different versions of PnP Modern Search (also with different versions of MGT) on the same page. We could of course also add the current version number to the disambiguation but that seems like overkill or what do you think?
I have updated the PR with these changes and hardcoded a disambiguation string of "pnp-modern-search". Because of the hardcoded string I also removed both of the webpart settings for this.
This now also updates all mgt elements and style tags for them dynamically when rendering so you never have to take disambiguation into account when designing the layouts or columns handlebars templates, you just keep using the standard elements like mgt-person etc. This then required some refactoring of TemplateService and TemplateRenderer but nothing major, just moved around a few functions.

wobba: You won't have multiple v4 versions anyways on one page so makes sense. I'll look it over and merge tomorrow 👍 And really appreciate this one!

wobba:
Thank you and awesome work! Hopefully all toolkit issues should be gone :)

PR #3558, opened 2024-02-22
Enable sort option for external items Microsoft Search

kanade-saket:
Hi Team , sort enabled in property pane for external items for Microsoft Search , 
this is tested for external items .

from MS docs >>  https://learn.microsoft.com/en-us/graph/search-concept-sort >> Sorting is also supported for [external items](https://learn.microsoft.com/en-us/graph/api/resources/externalconnectors-externalitem). The [property](https://learn.microsoft.com/en-us/graph/api/resources/externalconnectors-property) of external items to be sorted on should be refinable in the search schema.


wobba:
Thank you

PR #3569, opened 2024-02-27
Fix for selecting vertical using query string

### LINKED ISSUE #3452: Verticals tab not indicating selected value when using ?v=TabName or ?v=TabValue
### Discussed in https://github.com/microsoft-search/pnp-modern-search/discussions/3449

<div type='discussions-op-text'>

<sup>Originally posted by **boing0** January 14, 2024</sup>
I think I've got verticals set up correctly -- they refine the results as expected when I click on the tabs. When I use the query string v=TabName the results are also pre-refined for that tab value. But the selected tab doesn't indicate the query string value with the blue underline, it always shows the first tab as selected. I'm using {searchTerms} AND Site.Id:{verticals.value} in the Results Query template. I have verticals configured not as hyperlinks, which I thought was correct? I also tried v=TabValue (the Site.Id value) but that didn't work. Any idea what I'm doing wrong? I'm using 4.10.1.

Query string in use, results are correct but the right tab isn't selected, it's showing the first tab.
image

image

</div>

patrikhellgren:
Fixes #3452

### DIFF
--- a/search-parts/src/webparts/searchVerticals/components/SearchVerticalsContainer.tsx
+++ b/search-parts/src/webparts/searchVerticals/components/SearchVerticalsContainer.tsx
@@ -17,7 +17,7 @@ export default class SearchVerticalsContainer extends React.Component<ISearchVer
     super(props);
 
     this.state = {
-      selectedKey: undefined
+      selectedKey: props.defaultSelectedKey
     };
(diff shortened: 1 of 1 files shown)


wobba:
Thank you.

PR #3581, opened 2024-02-27
Fast-serve update to match the most recent changes.

s-KaiNet:
I noticed that you already use the latest (1.18.10) fast-serve version for SPFx 1.18 (hooray!). 
This PR simply adds some small updates regarding the usage of the latest fast-serve and other minor fixes: 

- `fast-serve/config.json` - not needed anymore, default config settings are used
- package.json's `serve` command was simplified for easier updates in the future
- `gulpfile.js` - `build.tslintCmd.enabled = false;` is not needed anymore, since tslint is not used in SPFx
- `fast-serve/webpack.extend.js` now fully matches the corresponding "configureWebpack" changes from gulpfile.js
- inside `config/serve.json` the "api" property is not used anymore in SPFx 1.18, but produces some errors for OOB "gulp serve", thus was removed


s-KaiNet:
I also found, that condition ` include: [/spfx-controls-react\/lib\/controls\/HoverReactionsBar/],` inside gulpfile.js doesn't work for me and build still produces unnecessary heavy output bundle. Most likely the problem is in windows styled paths, where back slashes are used. So I updated condition to match both windows and unix styles for `include`: 
obraz

search-parts/gulpfile.js
```diff
@@ -94,7 +87,7 @@ const envCheck = build.subTask('environmentCheck', (gulp, config, done) => {
             }, {
                 // Skip HoverReactionsBar from spfx controls as it's not used and is bundles
                 test: /index\.js$/,
-                include: [/spfx-controls-react\/lib\/controls\/HoverReactionsBar/],
+                include: [/spfx-controls-react[/\\]lib[/\\]controls[/\\]HoverReactionsBar/],
```
wobba, line 90: Windows 😅

wobba:
Thank you :)

PR #3606, opened 2024-03-06
New feature: Download files and folders from details list layout

patrikhellgren:
Adds a new feature to show a download button in the details list layout. You can download single files as is or multiple files and folders as a single zip file like in SharePoint document libraries.

Should also resolve the request in #3268.


search-parts/src/components/DownloadSelectedItemsButtonComponent.tsx
```diff
@@ -0,0 +1,195 @@
...
+        this.forceUpdate();
+      }
+    }
+
+    private _downloadSelectedItems() {
+
+      this.setState({
+        exporting: true
+      });
+      console.log(this._selectedItems);
```
wobba, line 88: Don't like console.log statements. Use the spfx logger or remove 🙂

patrikhellgren:
Removed console log statement and changed error logging.

kasperbolarsen:
Are you 100% sure that both selected items are documents? Can you share the query?

Issue #3617, opened 2024-03-11
add MSTEAMS protocol handler to support links to the new Teams Client


till-llit:
use case: People Search with Teams integration (direct call from search result)

currently only "tel" and "callto" are allowed url protocol handler to initiate calls from search results.

both are not working with the new Teams clients and result in error messages in new teams.

example URI like `tel://+08005555`  or `callto://first.last@domain.com` will result in error messages:

![image](https://github.com/microsoft-search/pnp-modern-search/assets/91128746/811ab55d-7045-42be-b02e-4cb93df41c73)

with msteams:// url is it possible to engage a direct call from a people search result.
example url: 

`msteams:/l/call/0/0?users=4:+08005555 `or `msteams:/l/call/0/0?users=first.last@domain.com` will initiate a direct call in MS Teams.

i will add a pull request to add msteams to the list of allowed URL protocols for search results.

wobba:
Merged and will be part of the next release.

PR #3618, opened 2024-03-11
add msteams to list of allowed uri protocols

till-llit:
see https://github.com/microsoft-search/pnp-modern-search/issues/3617


wobba:
Thank you

Issue #3619, opened 2024-03-11
Search Results - Details List: empty values are displayed now as "null" instead of empty


till-llit:
**Version used**
4.11.0

**Describe the bug**
search result with detailslist will show now the text value "null" in fields with empty values. 

**To Reproduce**

**Expected behavior**
empty values should be displayed as a empty string

patrikhellgren:
@till-llit I have submitted a PR for this now

RWordAZT:
Also noticed this in the employee directory I created on our Intranet. Would much rather see empty strings that "null", it clogs up the results with unnecessary text.

wobba:
Closing this issue. Please open a new one for other templates.

PR #3620, opened 2024-03-11
Render empty string instead of null and undefined values

patrikhellgren:
Fix for #3619


patrikhellgren:
Changed operator.

RWordAZT:
How do we implement the fix? My employee directory is still showing null values.
Null Values

patrikhellgren:
@RWordAZT It will be part of the next release. Waiting for that you could always format the column using handlebars so that if the value is 'null' then render an empty string instead.

Issue #3656, opened 2024-03-27
Query string for filters gets removed after clicking on Clear button


VictorRomanov1986:
**Version used**
Ex: 4.10.2

**Describe the bug**
In our solution we use custom page that includes Search Results and Search Filters webpart from PnP Modern Search package.
One of the ways how users can access this page is to navigate it from another page by clicking a link, that contains query string to preselect filters.
If the user wants to clear one of the filters and click Clear button
![image](https://github.com/microsoft-search/pnp-modern-search/assets/42931275/155a03d9-dc1b-42aa-8697-92cec8f66b56)
In this case the query string will be removed from the address, but all other filters will remain preselected 
![image](https://github.com/microsoft-search/pnp-modern-search/assets/42931275/2eee1700-c41f-4ed0-b1fd-fe02ade59337)
It means, that after refreshing the page all the filters will not be applied anymore
This is not happening if you simply click on cross icon near the selected filter
![image](https://github.com/microsoft-search/pnp-modern-search/assets/42931275/8a48f269-887c-4812-9ff7-c053a48d2d6d)

**To Reproduce**
Detailed steps to reproduce the behavior:
1. Go to page with Search Results and Search Filters webparts and with query string to preselect filters
2. Click on 'Clear' button for one of the filters
3. Query string will be removed

**Expected behavior**
Query String remains in the address bar if there are preselected filters

wobba:
Would be good if you can verify if the issue is in 4.11 as well as we don't bug fix on old releases.

VRomanovTau:
The issue is still reproducible in 4.11
I think the main problem is here - https://github.com/microsoft-search/pnp-modern-search/blob/71a208627bd6b9d192f8803fa40b41693a7216bf/search-parts/src/webparts/searchFilters/components/SearchFiltersContainer.tsx#L528C7-L528C35

PR #3671, opened 2024-04-04
Do not clear query string if there are applied filters

VictorRomanov1986:
Fixing #3656
Check whether there are applied filters. If no - remove query string. If yes - update query string


wobba:
Thank you!

PR #3675, opened 2024-04-08
Proper support of query template in Microsoft Search

lebiku:
Both 'queryString' and 'queryTemplate' in search query request contain the same data. This leads to wrong result sets. Graph Api now supports both properties with API v1.0. If correctly implemented, the result set is in accordance to search in graph explorer.

image

Having fixed the token replacement, the query part looks as follows:

image

see also Graph API [docu ](https://learn.microsoft.com/en-us/graph/api/search-query?view=graph-rest-1.0&tabs=http#example-2-basic-call-to-use-querytemplate)


wobba:
Thank you...and we should actually keep {searchTerms} in the query template for the SharePoint API as well I think. Haven't looked at this closely in a long time. Care to make it generic for both API's before I merge?

lebiku:
Hi @wobba, thank you. I was not sure of impacts / implementation details, so I kept it local. I changed the code to make it generic. In my tests it worked both, for Graph and SharePoint Search. Please feel free to change in case I didn't understand the whole picture and did the change in the wrong spot. 

Graph query after changes

image

Sharepoint query after changes

image

wobba:
Looks good!

PR #3715, opened 2024-04-25
Fix download when files are in a folder

patrikhellgren:
Fixes an issue that occured when downloading multiple files that are located in a folder.


wobba:
Awesome!

PR #3772, opened 2024-06-01
Add audience targeting to verticals

patrikhellgren:
Adds audience targeting for verticals. You can configure one or more Entra ID groups that the vertical should be visible for. If left empty the vertical is visible for all users.

Partially addresses #2052.


wobba:
Thank you!

PR #3809, opened 2024-06-18
Fix for enabling results grouping for newly added results webpart

patrikhellgren:
Fixes an issue that would occur when enabling results grouping in the details list layout for a newly added results webpart. After adding a results webpart and switching to details list layout the grouping toggle was did not change state until you first exited webpart properties and then edited it again.

### DIFF
--- a/search-extensibility/src/models/layouts/ILayout.ts
+++ b/search-extensibility/src/models/layouts/ILayout.ts
@@ -17,7 +17,7 @@ export interface ILayout {
     /**
     * Check if web part is in display or edit mode
     */
-    editMode: boolean
+    editMode: boolean;
 
     /**
      * Method called during the Web Part initialization.
--- a/search-parts/package-lock.json
+++ b/search-parts/package-lock.json
@@ -28,7 +28,7 @@
                 "@microsoft/sp-property-pane": "1.20.0",
                 "@microsoft/sp-webpart-base": "1.20.0",
                 "@pnp/common": "2.0.13",
-                "@pnp/modern-search-extensibility": "1.20.0",
+                "@pnp/modern-search-extensibility": "1.20.1",
                 "@pnp/spfx-controls-react": "3.20.0",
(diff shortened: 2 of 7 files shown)


wobba:
I think some improvement can be made here as your change will always load the grouping part - regardless of being used or not. Could it be moved to getPropertyPaneFieldsConfiguration?

patrikhellgren:
@wobba Sorry, had missed your earlier comment. I will have a look later today. If I remember correctly there is no way of passing the webpart display mode to the layout and there were existing lazy loadings without conditions so that's why it was done that way. I could add the display mode to the layout interface to have a good way of checking that in the layout. Any thoughts on that?

patrikhellgren:
@wobba I have now checked the code and I remembered correctly, there was already another lazy loading in place without any condition. I agree that it would be optimal if the lazy loading only happened if the webpart was in edit mode. I think the best option would be to add an editMode property to BaseLayout just like BaseDataSource already has and then pass this from the webpart. What is your opinion?

wobba:
@patrikhellgren I trust your judgement on passing the value :)

patrikhellgren:
@wobba I totally forgot about this one but now it should lazy load the property controls only if in edit mode.
Since I needed to update search-extensibility also and that had already been updated to SPFx 1.19 I guess this will not build until search-parts also is updated to 1.19 due to type mismatches. What is the plan there?

wobba:
Should we proceed with this one, and I can update extensibility as needed with an update to the 1.20 version.

wobba:
@patrikhellgren I published @pnp/modern-search-extensibility@1.20.1. so you can remove your extensibility changes, take a dep on this version and update once you resolve your merge conflicts (which I probably introduced)

patrikhellgren:
@wobba I have now bumped modern-search-extensibility dependency to 1.20.1 so this should be good to go.

PR #3810, opened 2024-06-18
Fix for details list selection and focus

patrikhellgren:
Fixes an issue that occurs when you have enabled multi-selection for the details list layout. When you have many result items so that the list covers more than the browser view, select the first item then scroll down so the first item is scrolled out of view, try to select another item and the view is scrolled to the top of the list and all items are de-selected. This fix removes setting the focus so that the selection is retained and the view is not scrolled to the top.

### DIFF
--- a/search-parts/src/components/DetailsListComponent.tsx
+++ b/search-parts/src/components/DetailsListComponent.tsx
@@ -430,13 +430,6 @@ export class DetailsListComponent extends React.Component<IDetailsListComponentP
 
         const { columns, items } = this.state;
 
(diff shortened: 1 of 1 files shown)


wobba:
Shouldn't the logic be kept as is if there is only one item selected?

patrikhellgren:
My testing shows that single item selection had the same behavior, when the selected item is outside of the visible area in the browser and you try to select another item the browser scrolls to the already selected one and all items are de-selected. This change resolves the behavior for both single and multi-select.

PR #3829, opened 2024-06-27
Show warning-message in property pane if there are slots missing (#3672)

jblueml:
This addresses [#3672 ](url).

If there are any slots missing (or not configured) but used in layout templates or result-type templates, a warning message will be displayed.

image


wobba:
Awesome addition!

PR #3834, opened 2024-07-02
Updates for German strings for Filter operators

VictorRomanov1986:
Updates for German strings for Filter operators


wobba:
Thank you :)

search-parts/src/loc/de-de.js
```diff
@@ -356,11 +356,11 @@ define([], function () {
             ApplyAllFiltersButtonLabel: "Anwenden",
             ClearAllFiltersButtonLabel: "Zurücksetzen",
             FilterNoValuesMessage: "Keine Werte für diesen Filter",
-            OrOperator: "OR",
-            AndOperator: "AND",
+            OrOperator: "UND",
```
JungLSEDV, line 359: i think there was a switchup between OR and AND, because OR means ODER and AND means UND, not the other way around.
same obv. for the other cases

Issue #3850, opened 2024-07-08
Search Results: Hide this web part if there's nothing to show: removes padding and margins


Lou-i3:
### Follow the template below or we will have to delete this request!!

**Version used**
Ex: 4.12.2

**Describe the bug**
On the PnP Search Result web part, when the "Hide this web part if there's nothing to show" parameter is enabled, the padding and margin of the web part are broken. 
When this setting is enabled on the WebPart, inspecting the DOM reveals that the div with data-automation-id="CanvasControl" wrapping the results has an inline style set to: `padding-top: 0px; padding-bottom: 0px; margin-top: 0px; margin-bottom: 0px;`. 

**To Reproduce**
Detailed steps to reproduce the behavior:
1. Add PnP Search Result web part to a page
2. Make sure there are results returned from the request
3. Enable "Hide this web part if there's nothing to show"
4. Publish the page 

**Expected behavior**
The layout of the web part should be the same wether this parameter is on or off. 

**Screenshots**
See screenshot, comparing with the setting off (left) and on (right): 
![image](https://github.com/microsoft-search/pnp-modern-search/assets/67910996/f091d68e-a1fc-4581-834c-be0e258cecaf)

**Desktop (please complete the following information):**
 - Browser: Brave 

**Additional context**
I tried with different requests and different templates but always get the same behaviour.

kasperbolarsen:
Hi,
Which layout are you referring to, or is it the same for all layouts?

Lou-i3:
Hi, 
Yes it is the same for all layouts, even Adaptive Cards.

wobba:
I suspect we need to add a check if a web part header is set - then don't remove padding.

wobba:
@Lou-i3 if you have a web part title, and there is no content, then the web part title should also be hidden - would that be the correct fix?

Lou-i3:
@wobba yes if there is no content, the web part should be completely hidden, title hidden too. 

My screenshot showed a lot of results because I just made an example for the screenshot. 
My scenario is that there aren't always documents from the query and if there are no documents, there shouldn't be anything displayed.

jblueml:
@wobba  I just checked this and figured out, what behaviour @Lou-i3 means: 
If you have 2 Search Results web parts side-by-side (e.g. in a 2-column-section) and _one of them_ has the setting "Hide this web part if there's nothing to show" set to TRUE, padding and margin are set to 0px - even if there are results to be shown. 
[...]

PR #3867, opened 2024-07-17
Fix for removing padding and margin if nothing to show

jblueml:
This solves #3850

### DIFF
--- a/search-parts/src/webparts/searchResults/components/SearchResultsContainer.tsx
+++ b/search-parts/src/webparts/searchResults/components/SearchResultsContainer.tsx
@@ -144,7 +144,7 @@ export default class SearchResultsContainer extends React.Component<ISearchResul
                 }
             }
             
-            // call handler if results found
+            // call handler if no results found
             this.props.onNoResultsFound();
         }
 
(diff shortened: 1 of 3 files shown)


search-parts/src/webparts/searchResults/components/SearchResultsContainer.tsx
```diff
@@ -143,6 +143,9 @@ export default class SearchResultsContainer extends React.Component<ISearchResul
                     renderTemplate = null;
                 }
             }
+            
+            // call handler if results found
```
wobba, line 147: "if no results are found" right?  Wrong comment text?

search-parts/src/webparts/searchResults/components/SearchResultsContainer.tsx
```diff
@@ -143,6 +143,9 @@ export default class SearchResultsContainer extends React.Component<ISearchResul
...
+            this.props.onNoResultsFound();
```
wobba, line 148: Should the comment read "..if no results found"?

jblueml: Your're right - thank you. I just committed the correction

PR #3875, opened 2024-07-22
Fix for displaying Details List sort arrows on multiple columns

### LINKED ISSUE #3839: Details list with multiple sorting columns - arrows on all columns change direction when one is clicked
**Version used**
Ex: 4.9.3 - updated to 4.12.2

**Describe the bug**
When using the details list layout with multiple sortable columns, clicking one column to sort changes the arrow direction on all columns.

**To Reproduce**
1. Set up a search result web part, make sure to pull in at least 2 fields based on (sortable) managed properties and refer to those in the layout slots configuration.
2. Edit Sort settings. Add the sortable managed properties as sortable fields, and check the "user sort" option on both.
3. Select the Details List layout. Open the Managed columns panel for the details list layout and add the 2 columns for the sortable managed properties. Check the "sortable" option and select the corresponding sort field.
4. At this point, the search results layout should show results in the details list layout with sortable columns.
5. Click on one of the column headers to sort. Click again to reverse the sort. BOTH column header arrows will change direction.

**Expected behavior**
The current experience is confusing as it's not clear which column is being sorted and all column header arrows change direction. I would expect the active arrow to be a different color to provide a visual cue that I'm sorting by that column, and I would expect other columns to retain their original arrow direction or have a different icon (such as an arrow that is both up and down) to indicate that sorting is possible but not actively sorted.

**Screenshots**
Initial state of configured web part is default sorted by PnPRevision date descending, although there is no visual indication:
image

Clicking "Site" sorts on that column but reverses ALL the arrows in all the columns:
image

 - Browser: Chrome

jblueml:
This should fix #3839.

As sorting by the user is only possible for one column at a time, a sort arrow is only displayed in the column according to which the search result is currently sorted.

### DIFF
--- a/search-parts/src/components/DetailsListComponent.tsx
+++ b/search-parts/src/components/DetailsListComponent.tsx
@@ -334,7 +334,8 @@ export class DetailsListComponent extends React.Component<IDetailsListComponentP
                         isRowHeader: false,
                         isResizable: column.isResizable === true,
                         isMultiline: column.isMultiline === true,
-                        isSorted: allowSorting,
+                        // show sort arrow only if the column is sortable and sorting has been applied by a previous click ('user sort' happened)
+                        isSorted: allowSorting && column.valueSorting === this.props.defaultSelectedField,
                         isSortedDescending: this.props.defaultDirection === SortFieldDirection.Descending,
                         sortAscendingAriaLabel: 'Sorted A to Z',
                         sortDescendingAriaLabel: 'Sorted Z to A',


wobba:
Thank you 🙂

corriespondent:
Thank you @jblueml!

kasperbolarsen:
in this case the results are sorted by Title (asc) as per the config but it does show. Would it be possible to show the initial sort as well ?

image


image

kasperbolarsen:
Yeah, that looks like a hard nut to crack

PR #3964, opened 2024-09-05
Update templating.md

Lou-i3:
Change theme variable name in documentation. 
To use theme variables we need to use "theme" and not "themeVariant".


wobba:
Thank you I trust you checked this :)

PR #3997, opened 2024-09-19
new scenario search-verticals-with-parameter

davemehr:
added the new search scenario search-verticals-with-parameter


wobba:
Thank you!

PR #3998, opened 2024-09-19
replace refinableyesno token

friendbool:
See https://github.com/SharePoint/sp-dev-docs/issues/9098
When using a YesNo type refiner, the tokens ǂǂ54727565 and ǂǂ46616c7365 do not work. In this case, I replace the tokens with the values true and false. Feel free to modify my code 👋

### DIFF
--- a/search-parts/src/helpers/DataFilterHelper.ts
+++ b/search-parts/src/helpers/DataFilterHelper.ts
@@ -126,6 +126,10 @@ export class DataFilterHelper {
                         value = "string('')";
                     }
 
+                    if (/RefinableYesNo/.test(filter.filterName)) {
+                        value = DataFilterHelper.fixRefinableYesNoFilter(filter, value);
+                    }
+
                     // Enclose the expression with quotes if the value contains spaces, or number only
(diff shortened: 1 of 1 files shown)


wobba:
This is interesting indeed and not really expected. Replacing with the strings are an ok workaround for sure. If the same issue also applies to OOB YesNo properties, then I'm wondering if we should rather test on the hex value and not on the property name?

friendbool:
Since the issue appears to be related to the refinement of YesNo fields, in SharePoint Online, the only YesNo fields enabled for refinement are those named RefinableYesNoXX. Therefore, we perform a double check: first on the field name and then on the field value. If you believe it would be better to remove the check on the field name, please proceed. Have a good evening, bye.

wobba:
There are a few other oob ones, but likely not important. I'll merge this for now and see if a fix can be done in the backend as the issue applies to the Graph API as well.

wobba:
@friendbool can you log this as an API issue on Graph so Microsoft gets a customer reporting it?

wobba:
Decided to merge this one :)

PR #4010, opened 2024-09-27
Token support for cookies

erkanch:
Added token support for cookies. Also updated the documentation for the same.


wobba:
Not sure I want to include this. This means you could read and display any cookie on the domain, and cookies are not really part of any official SharePoint API.

Issue #3934, opened 2024-08-23
PnP Search Filters v4 - 0 Results when filtering on multiple filter fields for external data source results


PinnuDivya:
**Version used**
 4.12.2.0

**Describe the bug**
I have created external data source connector to pull data from cosmos DB in M365 search admin center.
I am using microsoft search option as datasource and external Items as entity type. 
![image](https://github.com/user-attachments/assets/9b2301e5-3c18-4bb7-9577-22339897bc0c)

I am able to see the results in results webpart. Please see below for debug code.
![image](https://github.com/user-attachments/assets/7e1beb06-3f93-4724-9027-890c5fbec565)

I have also configured filters having 3 filter types. I have selected OR between the filters.
![image](https://github.com/user-attachments/assets/08b3a296-85e9-4f09-80bf-eaed36bc7db6).

The issue I am facing is when I select any 1 filter, I am able to see the results. But I see zero results when I select 2 or 3 filters.
Below image refers when one filter is been selected. When this filter is selected, past 24 hours in Modified Date filter is having 16 count.
![image](https://github.com/user-attachments/assets/d10033b8-191c-4281-b851-959f2f4e4a03)

When I select past 24 hours in Modified Date filter, results is showing as 0 instead of 16.
![image](https://github.com/user-attachments/assets/a1bf5b88-62cc-46ee-9626-59e1dd35cd2d)

And below is the query I am seeing when I have selected 2 filters.
f=[{"filterName":"filetype","values":[{"name":"docx","value":"\"ǂǂ6543218\"","operator":0,"disabled":false}],"operator":"or"},{"filterName":"lastModifiedTime","values":[{"name":"Past 24 hours","value":"2024-08-23T11:50:06.315Z","operator":5},{"name":"Past 24 hours","value":"2024-08-22T11:50:06.315Z","operator":4}],"operator":"or"}]#

**To Reproduce**
Steps to reproduce:
1. Create external data source connector in M365 search admin center
2. Configure PnP search results for external data entity type and configure filters on any sharepoint page.
3. Select 2 or more filters.

**Expected behavior**
User should see the results in results webpart.

**Desktop (please complete the following information):**
 - Browser : Tested in Edge

**Additional context**
I have implemented same with out PnP and filters worked as expected
<img width="513" alt="image" src="https://github.com/user-attachments/assets/84daae85-0e41-4d7b-8bc8-a68dbb897119">

It would be very helpful if any one could help me here.

[...]

PR #4087, opened 2024-11-07
Using KQl to filter search results in microsoft search

akshataggrwal:
Using KQL to filter search results while using microsoft search instead of aggregation filters.

Issue Reference - https://github.com/microsoft-search/pnp-modern-search/issues/3934


wobba:
I think we should still send as FQL, but add multiple rows with one row per each. At least for the AND clause. And a PR which comments out lines of code is not good.

wobba:
Created #4090 which also support dates in KQL. I simplified and use `..` for date ranges, which means GE and LT for start and end date.

akshataggrwal:
Hi @wobba - Thanks for the updates. I tested the changes in your [PR(4090)](https://github.com/microsoft-search/pnp-modern-search/pull/4090) locally for 'externalItem' entity type and it is working as expected.

Is it reasonable to assume for this to available in next release?

wobba:
Thank you for test @akshataggrwal. I'll merge it and get a release going.

PR #4118, opened 2024-11-22
Add null check for PrimaryQueryResult in SharePointSearchService

Niels-PuurICT:
Added a null check to the searchResponse.PrimaryQueryResult because Selected properties doesn't load on some sites because it's null. (PnP - Search Results)


wobba:
Thank you :)

Issue #4121, opened 2024-11-22
Filters with RefinableYesNo Managed Properties


sympmarc:
**Version used**
Ex: 4.14.0

**Describe the bug**
When using a RefinableYes/No managed property as a filter, there are two issues:
* The display of the filter shows False and True instead of No or Yes
* Apply a filter (choosing True or False) returns no results.

**To Reproduce**
Detailed steps to reproduce the behavior:

1. Map a Yes/No crawled property to one of the RefinableYesNo Managed Properties (I tried both the `ows_IsPolicy` and `OWS_Q_BOOL_ISPOLICY` options)
2. Add a filter using the RefinableYes/No MP
3. Try filtering
 
**Expected behavior**
Filtering should work

**Screenshots**

Filter
![image](https://github.com/user-attachments/assets/0baa0c51-d133-491e-ae43-f557dbd27213)

Results
![image](https://github.com/user-attachments/assets/d176f71c-5be4-4830-9890-ef4be54b92df)

**Desktop (please complete the following information):**
 - Edge

kasperbolarsen:
Bug confirmed

wobba:
This relates to https://github.com/microsoft-search/pnp-modern-search/pull/3998. As it's actually a data issue and not web part issue I'm note sure I want to fix it here. I did reach out to the service team but haven't heard anything back, but will ping them.

sympmarc:
I only last week figured out those RefinableYesNoxx MPs existing. Should it work properly with RefinableString? (I can't remember ever trying with Yes/No columns.)

wobba:
@sympmarc string should work.. and check what the actual string value is.

kasperbolarsen:
@sympmarc  can you check your findings again using v.4.16? I get the true/false texts in the Refiner, but they works as expected for me.

sympmarc:
@kasperbolarsen - Which do you want me to test? The RefinableYesNo MP or RefinableString?

kasperbolarsen:
RefinableYesNo

sympmarc:
Yes, it seems to work in 4.16. Was it fixed in #3998 ?

It still says True/False, but the filtering seems to work. Given it's a Yes/No, it feels like it should use those strings instead.

kasperbolarsen:
I am working on localized values for the supported languages

PR #4239, opened 2025-02-28
Enhance boolean value localization in SharePointSearchDataSource and …

kasperbolarsen:
…add translation method in LocalizationHelper #4121


search-parts/src/dataSources/SharePointSearchDataSource.ts
```diff
@@ -1086,8 +1086,18 @@ export class SharePointSearchDataSource extends BaseDataSource<ISharePointSearch
...
                         }
                     });
-                } else {
-                    updatedValues.push(value);
+                } else 
+                {
+                    if(filterResult.filterName.toString().indexOf("RefinableYesNo")>-1) 
+                    {
+                        value.name = LocalizationHelper.getTranslatedBooleanValue(value.name, lcid);
+                        updatedValues.push(value);
```
wobba, line 1094: Can you rather do something like this and get the strings into the loc files?

export class LocalizationHelper {

    /**
     * Gets the localized string for the given key and locale.
     * @param key The key of the string to localize.
     * @param locale The locale to use for localization.
     */
    public static async getLocalizedString(key: string, locale: string): Promise<string> {
        try {
            const strings = await import(`../loc/${locale}.js`);
            return strings.default[key] || key;
        } catch (error) {
            console.error(`Error fetching localized string for locale ${locale}:`, error);
            return key; // Return the key if no localization found
        }
    }
}

import { LocalizationHelper } from '../helpers/LocalizationHelper';

private async _getLocalizedStringExample() {
    const key = "HelloWorld";
    const locale = "fr-FR"; // Example locale
    const localizedString = await LocalizationHelper.getLocalizedString(key, locale);
    console.log(localizedString); // Output the localized string
}

kasperbolarsen: You are pushing it, I had a hard time just getting the current version to work ;-)

PR #4252, opened 2025-03-11
Added Czech localization

sliziky:
Added Czech localization in `cs-cz` file as discussed in here [#4249](https://github.com/microsoft-search/pnp-modern-search/discussions/4249)


sliziky:
You're right, somehow missed that. Will do :)

wobba:
Thank you!

Issue #4254, opened 2025-03-12
Tokens CurrentHourUTC, CurrentMinuteUTC and CurrentSecondUTC returning 1 digit values results in no search results


EMaarten:
**Version used**
4.16.0

**Describe the bug**
The tokens CurrentHourUTC, CurrentMinuteUTC and CurrentSecondUTC can return 1 digit values depending on the time. This results in a search query returning no results. The tokens should return 2 digit values so leading zeros should be added when needed. 

For the tokens CurrentDateUTC and CurrentMonthUTC 1 digit values don't seem to be a problem.

**To Reproduce**
Detailed steps to reproduce the behavior:
1. Add a search results web part
2. Select SharePoint a data source
3. Configure the search query e.g. Created<="2025-03-12T08:07:06Z"
4. Click the button to apply the search query
5. Check the search results (items are returned)
6. Now change the search query to  _Created<="2025-3-12T{CurrentDateUTC}:{CurrentMinuteUTC}:{CurrentSecondUTC}Z"_
7. Click the button to apply the search query
8. Make sure the query gets executed when the current date, current minute or current second is in the range 0-9
9. Check the search results (no items are returned)
10. Manually change the search query to the current time but with leading zeros
11. Click the button to apply the search query
12. Check the search results (items are returned)

**Expected behavior**
The tokens CurrentDateUTC, CurrentMinuteUTC and CurrentSecondUTC should return 2 digit values, so leading zeros should be added when needed. This way the search query keeps returning search results.

**Screenshots**

**Desktop (please complete the following information):**
 - Browser Chrome

![Image](https://github.com/user-attachments/assets/0da38e26-33bf-45e6-b61f-3b3aff93a957)

![Image](https://github.com/user-attachments/assets/c5e36348-82a2-4649-8860-d7ada47f7c20)

![Image](https://github.com/user-attachments/assets/8d8e82df-83b6-42f2-b439-abfe5ac17b7f)

![Image](https://github.com/user-attachments/assets/518df38b-77ca-4e24-b7af-24399c068b72)

**Additional context**
-

EMaarten:
There might be customer implementations based on the current format so maybe one could add an optional format parameter to the tokens?  This way one can search for the current tokens but also perform a search for the tokens with the new format parameter based on multiple regular expressions in the "replaceDateTokens" method in the TokenService? 

@wobba what would you advise?

wobba:
I'd say we add new tokens.

[...]

PR #4309, opened 2025-04-12
Adding new date tokens respecting 2 digits for issue #4254

diabhoil:
Added new tokens for the date replacement which will respect a leading zero for numbers < 10.

Referring to issue #4254


wobba:
@diabhoil any way you think looks good is ok :)

PR #4313, opened 2025-04-14
Add localized strings for mgt-person-card and its sub-components

patrikhellgren:
Resolves #1714 which also has the discussion about this.

Also update MGT to the latest version to handle the latest localization strings.


wobba:
Thank you!

PR #4317, opened 2025-04-15
Add compatibility matrix for extensibility projects

a1mery:
The idea behind this pull request is to list for all PnP Modern Search release, the SPFx version it uses so it is easier to know which version to use when developing extensibility projects (easier than looking directly at the source code).


wobba:
Wow...awesome work!

PR #4363, opened 2025-05-15
Update templating.md - fixing html error

Lschockaert:
Small typo in the html structure of the getAttachments helper.

The closing html tag containts href while the element is an '< a >'


wobba:
Thank you!

PR #4373, opened 2025-05-22
ensure extensions are always loaded

robi26:
When a custom data source is configured, the following error may occur: ``Cannot read properties of undefined (reading 'serviceKey').``
This error happens because the render method is called before the loadExtension has finished / is called. When attempting to load the data source in the render method, the source cannot be found:
```typescript 
// SearchResultsWebPart.ts:2108
const source = this.availableDataSourceDefinitions.find(definition => definition.key === dataSourceKey);
serviceKey = source.serviceKey; // << error happens here
```
This pull request ensures that the data source is available and extensions are loaded before attempting to access the serviceKey.


wobba:
Thank you!

Issue #4377, opened 2025-05-27
Search Results / Verticals webparts are not working in Viva Connections (Teams App)


VictorRomanov1986:
**Version used**
Ex: 4.16

**Describe the bug**
Search Results / Verticals webparts are not working in Viva Connections (Teams App) when user open it for the first time
![Image](https://github.com/user-attachments/assets/1a74dcef-198c-4b24-afa7-46be31d98328)

It happens when user type the text in Search box on top of Teams 
![Image](https://github.com/user-attachments/assets/ccc0d93c-9dcf-453a-9e08-73d977816529)

If user will click on Refresh button - webpart start to work

**To Reproduce**
Detailed steps to reproduce the behavior:
1. Go to 'Microsoft Teams App'
2. Click on 'Viva Connections' app that is configured with SharePoint home page
3. Type in search box any text, hit Enter

**Expected behavior**
PnP Search Results webparts loaded correctly

kasperbolarsen:
I am not sure we support that scenario, @wobba @patrikhellgren , any thoughts?

wobba:
@VictorRomanov1986 can you retest with a build from the dev branch?

VictorRomanov1986:
@wobba One bug I guess still exist - 
<img width="1213" height="545" alt="Image" src="https://github.com/user-attachments/assets/8fbf464e-2d05-4215-9614-f8289c674c42" />
The placeholder section is still on the page, despite data has been already loaded. No matter what I tried. This is our custom template 
<img width="823" height="436" alt="Image" src="https://github.com/user-attachments/assets/45689ba3-d9a3-472a-9ee4-c467c635ec21" />

Interesting thing, that built-in templates use different values for id`s
<img width="544" height="586" alt="Image" src="https://github.com/user-attachments/assets/5e09ff02-c4f3-4a06-b59d-f3e6fd46d7cd" />

VictorRomanov1986:
@wobba Any updates regarding my last comments?

PR #4391, opened 2025-06-02
Fix for #4377. +Reducing bundle size

VictorRomanov1986:
Fix for [#4377](https://github.com/VictorRomanov1986/pnp-modern-search/issues/4377).
Reducing bundle size by importing individual components instead of the whole package


wobba:
Can you validate the size change of doing this? Webpack has tree shaking so direct imports should not be needed since a long time ago.

wobba:
Wow.. I would have thought webpack did better than this.

PR #4497, opened 2025-10-23
Fix Microsoft Search filtering

patrikhellgren:
Fixes an issue when using filtering with Microsoft Search causing a ```'n is not a function'``` error.


wobba:
Thank you!

PR #4515, opened 2025-11-03
Download selected list items #4498

kasperbolarsen:
This PR makes it possible to export the values of selected items/rows in the Details List to a CSV file
Added feature, updated loc files and documentation

### DIFF
--- a/search-parts/src/components/DownloadSelectedItemsButtonComponent.tsx
+++ b/search-parts/src/components/DownloadSelectedItemsButtonComponent.tsx
@@ -79,28 +79,16 @@ export class DownloadSelectedItemsButtonComponent extends React.Component<IExpor
           return null;
         }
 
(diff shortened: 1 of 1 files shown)


search-parts/src/components/DownloadSelectedItemsButtonComponent.tsx
```diff
@@ -57,15 +64,56 @@ export class DownloadSelectedItemsButtonComponent extends React.Component<IExpor
...
+              disabled={
+                this.state.exporting ||
+                !requiredPropertiesAvailable
+              }
+              onClick={this._downloadSelectedItems}
+              theme={this.props.themeVariant as ITheme}
+            />
+          );
+        }
+
```
wobba: Seems the two if's can be refactored as there's a lot of shared code. Iconprops only which is different?

PR #4532, opened 2025-11-17
Enhance SharePoint Search Data Source with Preview Functionality

kasperbolarsen:
- Added methods to generate AutoPreviewUrl and AutoPreviewImageUrl for items in SharePointSearchDataSource.
- Implemented checks for online domains and M365 sources to validate preview URLs.
- Updated the getItemsPreview method to enhance item properties with preview information based on configured template slots.
- Modified default_cards.html and CardsLayout.ts to utilize new slots for SiteTitle and SPWebURL.
- Refactored SearchResultsContainer to call the new getItemsPreview method from the data source.
- Removed redundant preview logic from SearchResultsContainer.


estruyf:
One thing to note, @kasperbolarsen @wobba: this is a breaking change for customers using the extensibility model to implement their own data sources. I had a customer report that the images were no longer working on their results. While investigating it, I saw it was related to this, where Data Sources now have to implement their own image logic. Understandably, but as it was a minor release, they didn't expect any issues with the update.

Issue #3227, opened 2023-09-01
Selecting a filter affects multiple Search Results Web Parts


jacksonv1lle:
**Version used**
Ex: 4.9.3

**Describe the bug**
When there are two sets of Filters/Results on a page, selecting a filter from one of the Filter Web Parts will affect both Result Web Parts.

When selecting the 'Any time' value, it behaves as expected. I guess this is because the filters are being cleared in the URL query.

**To Reproduce**
1. Add 2x Search Results web part
2. Add 2x Search Filters web part
3. Connect the filter web parts to their corresponding results web parts
4. Edit both filter web parts to have a filter by 'Created' - Date interval type
5. Edit both results web parts to have a suitable results source
6. Save the page
7. Now when you select a value from one filter, both filters are updated

**Expected behavior**
When selecting a filter from one of the Filter Web Parts, it should only affect the corresponding Result Web Part.

**Screenshots**

https://github.com/microsoft-search/pnp-modern-search/assets/2648490/0a9c1530-df97-4923-99f2-57e307d7b5e1

**Desktop (please complete the following information):**
 - Chrome / Windows

**Additional context**
I believe this may be related to [issue 3130](https://github.com/microsoft-search/pnp-modern-search/issues/3130) but that bug seems to be describing an issue regarding verticals not clearing the filter values.

This bug was introduced with [PR #2164](https://github.com/microsoft-search/pnp-modern-search/pull/2164) to address Issue #2153 which was originally caused by the SharePoint router as @FranckyC [mentions here](https://github.com/microsoft-search/pnp-modern-search/issues/2153#issuecomment-1159536702) however I believe the SharePoint router now behaves differently. 

I have tried to reproduce issue #2153 after reverting the change from PR #2164 and each time I selected the navigation link the whole page is refreshed.

kasperbolarsen:
would you say this case is the same issue as in https://github.com/microsoft-search/pnp-modern-search/issues/3130  ?

hstuermannfellowmind:
We are having the same issue. 

This behavior comes - imo - from these lines in the SearchFiltersContainer.tsx. The component is listening to changes in the URL, and then pushes the filters to the connected search result filter webpart.

```typescript
 ((history) => {
      let pushState = history.pushState;
      history.pushState = (state, key, path) => {
        pushState.apply(history, [state, key, path]);
        this.getFiltersDeepLink();
      };
    })(window.history);
```

[...]

PR #4570, opened 2025-12-11
Ensuring that selecting a filter affects only the associated results web part

kasperbolarsen:
Selecting a filter affects multiple Search Results Web Parts #3227


kasperbolarsen:
Addressed comments from SonarQubeCloud, refactored offending 4 levels deep code

estruyf:
@kasperbolarsen @wobba just to note, these kind of changes are breaking the original experience. I was informed by a customer that had a "continue search" experience with filter set up, and with the new version, it doesn't work anymore. As the redirect to the search page was with the `?f=`, and now it requires `?f_<instance_id>=`.

It would have been "better" to have an option to change the behaviour. Do you want to have targetted filtering (`?f_<instance_id>=`), or general (`?f=`).

Issue #4649, opened 2026-02-12
Date range filter: page scrolls to top when selecting date via calendar icon


Jintow:
**Version used**
4.19.1.0

**Describe the bug**

In the Search Filters web part, the date range filter uses two date inputs (From/To). Each input has (1) a text field that opens a calendar picker when focused, and (2) a calendar icon on the right that also opens the picker.

The bug is that selecting a date via the text field behaves correctly, but selecting a date after opening the picker via the calendar icon causes the page to jump/scroll to the top immediately after the date selection. This is disruptive because the user loses their position in the page.

**To Reproduce**

1 - Add a Search Filters web part to a modern SharePoint page.

2 - Configure a filter as a date range filter (two inputs: From / To) based on the Created managed property.

3 - Publish the page and open it.

4 - Scroll down so the filter is not at the very top of the viewport (page is scrolled).

4 - In the From (or To) input, click the calendar icon (right side of the input) to open the date picker.

5 - Select any date from the calendar.

**Expected behavior**

The date is applied and the page keeps the current scroll position (no jump).

**Screenshots**

![Image](https://github.com/user-attachments/assets/ff3673b8-b412-44e3-8d35-e42cdf89ff5c)

**Desktop (please complete the following information):**
 - Chrome

**Additional context**

mobun:
I took a look at it today.
Root cause seems to be the Fluent UI DatePicker callout focus restore.
When the callout closes, Fluent UI restores focus via `focusAsync`/`element.focus()`. 
This causes the scroll container `data-automation-id="contentScrollRegion"` to jump to `scrollTop=0`, so the page scrolls to the top after date selection.

Fix idea: override focus restore via `calloutProps.popupProps.onRestoreFocus` and call `originalElement.focus({ preventScroll: true })` to keep accessibility behavior but avoid the scroll jump. I created a PR for that.

wobba:
Fixed.

PR #4652, opened 2026-02-14
Fix DatePicker scroll jump by preventing focus restore scroll

mobun:
* Prevent the DatePicker callout from restoring focus in a way that scrolls the SharePoint content container.
* Keeps focus behavior for accessibility while avoiding the page jump to top after date selection.

#4649


wobba:
Thank you

PR #4666, opened 2026-03-03
Fix thumbnail URL construction in DataSourceHelper

estruyf:
Fix in the `generateGraphThumbnailUrl`. There were two `thumbnails` in the path.


wobba:
Thank you!

PR #4711, opened 2026-04-06
fix: Ace editor cursor position offset in template dialog (#4706)

### LINKED ISSUE #4706: Template editor cursor position issue
**Version used**
4.19.1

**Describe the bug**
When editing templates using the provided editor, the cursor and text selection is showing as an offset of where the cursor actually is.

For example, if you double click on a word to select it, the selection looks like its on the side of the text, and there is no way of knowing where your cursor is - it is always displaying to the right of the actual position.

**To Reproduce**
1. add a search results webpart
2. switch the property panel to the layouts page and select "custom"
3. click on the {} icon to open the template editor
4. use the mouse to place the cursor somewhere (middle of word for example) and start typing
5. See that what you type is not being entered in the position where you placed the cursor

**Expected behavior**
Editor should place the cursor where the mouse clicks, and display a correct cursor indicator and text selection

 - Browser : Edge

I could not find an existing issue with this which is surprising...

wobba:
## Problem

Fixes #4706 — When editing templates using the built-in code editor, the cursor and text selection appear at an offset from where the user clicks.

## Root cause

The Fluent UI `Dialog` component uses CSS `transform: translate(...)` for centering. Ace editor calculates mouse-to-text coordinate mapping based on DOM positions, but CSS transforms shift the visual rendering without changing the layout coordinates that Ace reads. This causes clicks to register at the wrong text position.

## Fix

Added an `onLoad` callback on the AceEditor component that calls `editor.resize()` via `requestAnimationFrame`. This forces Ace to recalculate its coordinate mapping after the dialog has fully rendered and its CSS transform has been applied.

Note: The issue reporter mentioned this only occurs on one tenant, suggesting it may be triggered by specific SharePoint CSS or layout variations. This fix handles it generically.

### DIFF
--- a/search-parts/src/controls/TextDialog/TextDialog.tsx
+++ b/search-parts/src/controls/TextDialog/TextDialog.tsx
@@ -114,6 +114,11 @@ export default class TextDialog extends React.Component<ITextDialogProps, ITextD
                                     $blockScrolling: Infinity
                                 }
                             }
+                            onLoad={(editor) => {
+                                // Force resize after the dialog's CSS transform is applied,
+                                // otherwise Ace miscalculates cursor/click positions
+                                requestAnimationFrame(() => editor.resize());
+                            }}
                             name="CodeEditor"
                             enableBasicAutocompletion={true}
                         />


ishaisagi-hns:
lovely!

PR #4800, opened 2026-05-28
Fix Webpart title styles

jblueml:
Fixes a small bug setting the font-size of the Webpart title as discussed [here](https://github.com/microsoft-search/pnp-modern-search/discussions/4621#discussioncomment-17064006).


wobba:
Thank you!

PR #4853, opened 2026-07-06
fix(templating): restore CSS nesting in template styles (#4851)

### LINKED ISSUE #4851: CSS Style Nesting not working as expected in version 4.23.1
**Version used**
4.23.1

**Describe the bug**
I have lots of advanced html Templates that I use since PnP 4.14 and with the newest Update test most of them broke from a design perspective which I deemed super strange at first glance. After investigating I found that the CSS Nesting behaviour turned inconsistent with the newest version. 

A further explanation: I use a combination of prefixed styles (normal style blocks) and non-prefixed (via data-cssscope="layer") to create responsive behaviour with different section sizes in SharePoint. For better maintainability I use nested CSS styles for a while now, which always worked great.

When testing around I found out, that nestet styles are rendered partially (which is weird in its own way) or not at all in a template, except for when I add data-cssscope="layer" for ALL of the styling. As I often use the same template multiple times on a page with different section backgrounds I cannot un-prefix all of the styles, because the semantic colors are overwriting each other. So I deeply hope there will ne a fix for this new behaviour. I added a lightweight html template snippet that I used for testing below.

**To Reproduce**
1. Create page and add a PnP Modern Search Results webpart and get any result list to display
2. add this design template code (Codeblock feature seems to be behaving weirdly so I added the html file)

[test-case.html](https://github.com/user-attachments/files/29693091/test-case.html) 

**Expected behavior**
See Screenshot provided below. I expect it to look like in the right even without the data-cssscope="layer", but it looks like in the left (which is weird because the link styling is also nested but it gets interpreted while the rest (like icon size and margin) does not. Also take a look at the pagination - it looks different.

**Screenshots**
Image

 - Browser Chrome

wobba:
### Problem
Nested CSS inside Handlebars template `<style>` blocks stopped being applied in 4.23.1 (rendered partially or dropped), unless `data-cssscope="layer"` was used for *all* styles.

### Root cause
PR #4840 (commit `beeb526f`, shipped in 4.23.1) refactored `TemplateRenderer` and switched the non-MGT template parse from `DOMParser.parseFromString` (full Document → `style.sheet` populated) to `Range.createContextualFragment` (DocumentFragment → `style.sheet === null`). With no `sheet`, `legacyStyleParser` fell through to its regex fallback, whose property matcher `[^{}]*` cannot span nested `{}`, so nested rules were dropped or flattened.

### Fix
When `style.sheet` is unavailable, re-parse the CSS text through a throwaway **inert** `DOMParser` document (CSS assigned via `textContent`, so no markup breakout and scripts never run) to obtain a populated CSSOM sheet. This restores the CSSOM prefixing path, which preserves CSS nesting and `@media` rules. The regex fallback remains only as a last resort.

### Verification
- Confirmed root cause in a real browser: `DOMParser` → `sheet` present + nesting preserved in `cssText`; `createContextualFragment` → `sheet === null`.
- Verified end-to-end on a live Search Results web part with the reporter's exact template: the injected `<style>` now emits `#pnp-template_… .mytest { … & pnp-icon { font-size: 26px … } & li { … } }` and computed styles confirm the nested rules apply (icon 26px/10px, li flex/8px, link 500/18px).
- Security: a `</style>` injection test creates no element and does not execute.
- `heft build` passes (no TS/lint errors).

### DIFF
--- a/search-parts/src/services/templateService/TemplateService.ts
+++ b/search-parts/src/services/templateService/TemplateService.ts
@@ -313,13 +313,53 @@ export class TemplateService implements ITemplateService {
         await this._customElementHelperPromise;
     }
 
+    /**
+     * Re-parses raw CSS text through a throwaway, inert DOMParser document to
+     * obtain a populated CSSOM stylesheet. This is used when a style element is
+     * detached (e.g. created via Range.createContextualFragment) and therefore
+     * exposes a null `style.sheet`. Going through the browser's real CSS engine
+     * preserves modern features such as CSS nesting, which the regex fallback
+     * cannot handle. The document produced by DOMParser is inert (scripts never
+     * run) and the CSS text is assigned via textContent, so no markup breakout
+     * or code execution is possible.
+     */
+    private parseCssTextToSheet(cssText: string): CSSStyleSheet | null {
+        if (!cssText || !cssText.trim()) {
+            return null;
+        }
+
+        try {
+            const doc = new DOMParser().parseFromString("<style></style>", "text/html");
+            const styleEl = doc.getElementsByTagName("style")[0];
+
+            if (!styleEl) {
+                return null;
+            }
+
+            styleEl.textContent = cssText;
+            return (styleEl.sheet as CSSStyleSheet) ?? null;
+        } catch {
+            return null;
+        }
+    }
+
     public legacyStyleParser(
         style: HTMLStyleElement,
         elementPrefixId: string
     ): string {
         let prefixedStyles: string[] = [];
 
-        const sheet: any = style.sheet;
+        let sheet: any = style.sheet;
+
+        // When the style element is detached (e.g. created via
+        // Range.createContextualFragment), style.sheet is null and the CSSOM is
+        // unavailable. Re-parse the CSS text through a throwaway DOMParser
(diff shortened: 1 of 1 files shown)


wobba:
Addressed review feedback in 26a259d2:

- **SonarCloud `typescript:S6582`** — replaced `!cssText || !cssText.trim()` with an optional chain `!cssText?.trim()`.
- **Copilot (SecurityError on `cssRules`)** — added a `safeGetCssRules()` helper that wraps `sheet.cssRules` in try/catch and returns `null` on failure. Both the initial check and the CSSOM path now go through it, so a thrown `SecurityError` (e.g. a cross-origin stylesheet) degrades to the text-based fallback instead of aborting template rendering.
- **Copilot (throwaway `<style>` construction)** — the temp element is now created via `doc.createElement('style')` and appended to the **inert** DOMParser document's `head` (not the live document), avoiding an extra HTML parse + DOM traversal while still populating `style.sheet`. It stays inert, so no `@import`/`url()` fetches fire and the page is unaffected.

Re-verified end-to-end on a live Search Results web part: the injected `<style>` keeps full nesting and computed styles apply correctly (icon 26px/10px, li flex/8px, link 500/18px).

Issue #4891, opened 2026-07-29
BUG: Filter picker truncates refinement values containing hyphen and shows only prefix text


estruyf:
**Version used**
4.23.3

**Describe the bug**

When refinement values contain additional text separated by a hyphen, the filter picker only shows the first part of the label.

Example refinement values returned:

- AC Rotterdam - FACT 3
- AC Rotterdam - FACT 2
- AC Rotterdam - FACT 1

Displayed in the picker:

- AC Rotterdam
- AC Rotterdam
- AC Rotterdam

This makes different refinement options indistinguishable.

**To Reproduce**

- Configure a Search Filters web part with a filter bound to a managed property that returns string refinement values containing hyphen-separated suffixes.
- Ensure values like AC Rotterdam - FACT 1, AC Rotterdam - FACT 2, AC Rotterdam - FACT 3 are present in refinement results.
- Open the filter picker (for example ComboBox or similar picker-based filter template).
- Observe the displayed labels.

**Expected behavior**
The picker should display the full refinement label exactly as returned, for example:

- AC Rotterdam - FACT 3
- AC Rotterdam - FACT 2
- AC Rotterdam - FACT 1

**Screenshots**

<img width="449" height="99" alt="Image" src="https://github.com/user-attachments/assets/9008538b-35c2-488e-9f1d-b126608ae840" />

**Desktop (please complete the following information):**
 - Chrome

kasperbolarsen:
Available in version 4.24.0

PR #4892, opened 2026-07-29
Fix refiner labels with hyphen suffix

estruyf:
Fixes the hyphen issue in the refiner values. Related to issue: #4891


wobba:
Thank you!

